PR #2, opened 2020-11-04
Add installer role for running the foreman-installer

ehelms:
Starting with an initial role for the foreman_installer providing the most basic of functionality regarding the installer. There are many sources of inspiration for this role, and of varying complexity and handling. I wanted to start with the simplest, most re-usable form of that functionality.

I am starting this in Draft format to see what others think and also to try to figure out the appropriate level of testing for a role of this type.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -110,3 +110,32 @@ jobs:
         with:
           name: collection
           path: theforeman-operations-*.tar.gz
+
+  molecule:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        ansible:
+          - stable-2.9
+          - stable-2.10
+          - devel
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python
+        uses: actions/setup-python@v2
+        with:
+          python-version: "3.7"
+      - name: Install Ansible
+        run: pip install --upgrade git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
+      - name: Install dependencies
+        run: make test-setup
+      - name: Run tests
+        env:
+          DRIVER_NAME: docker
+        run: |
+          for roledir in roles/*/molecule; do
+            pushd $(dirname $roledir)
+            molecule test
+            popd
+          done
--- a/.yamllint
+++ b/.yamllint
@@ -1,0 +1,4 @@
+---
+rules:
+  line-length:
+    max: 140
--- a/Makefile
+++ b/Makefile
@@ -7,7 +7,7 @@ ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
 METADATA := galaxy.yml LICENSE README.md requirements.txt CHANGELOG.rst
 $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(eval _$(PLUGIN_TYPE) := $(filter-out %__init__.py,$(wildcard plugins/$(PLUGIN_TYPE)/*.py))))
-DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(wildcard $(ROLE)/*/*))
+DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(filter-out $(ROLE)/molecule/%, $(wildcard $(ROLE)/*/*)))
 
 PYTHON_VERSION = $(shell python -c 'import sys; print("{}.{}".format(sys.version_info.major, sys.version_info.minor))')
 COLLECTION_COMMAND ?= ansible-galaxy
--- a/roles/installer/README.md
+++ b/roles/installer/README.md
@@ -1,0 +1,32 @@
+theforeman.operations.installer
+===============================
+
+Run the foreman-installer
+
+Role Variables
+--------------
+
+Required:
+
+- `installer_scenario`: The installer scenario to run, this is required to be set
+
+Optional:
+
+- `installer_options`: Array of options to pass to the installer
+- `installer_verbose`: Enables verbose output mode in the installer
+- `installer_no_colors`: Disables color output from the installer
+- `installer_command`: Installer command to run, can be used by derivative projects to specify a branded command
+
+Example Playbooks
+-----------------
+
+Run the installer setting the initial admin password:
+
+```yaml
+- hosts: target-host
+  roles:
+    - role: theforeman.operations.installer
+      vars:
+        installer_options:
+          - '--foreman-initial-admin-password changeme'
+```
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -1,0 +1,29 @@
+---
+- name: Check that necessary variables are defined
+  assert:
+    that:
+      - installer_scenario is defined
+
+- name: Check if initial install has happened
+  stat:
+    path: /etc/foreman-installer/scenarios.d/.installed
+  register: installed
+
+- name: 'Check if changes are needed by running installer in noop mode'
+  command: >
+    {{ installer_command }}
+    --scenario {{ installer_scenario }}
+    --noop
+  register: installer_noop
+  changed_when: false
+  failed_when: installer_noop.rc not in [0, 2]
+  when: installed.stat.exists
+
+- name: 'Run installer'
+  command: >
+    {{ installer_command }}
+    --scenario {{ installer_scenario }}
+    {{ (installer_verbose|bool) | ternary("-v", "") }}
+    {{ (installer_no_colors|bool) | ternary("--no-colors", "") }}
+    {{ installer_options | join(' ') }}
+  when: installer_noop.rc is not defined or installer_noop.rc == 2
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,9 +1,3 @@
-pytest
-pytest-xdist
-vcrpy
-ansible_runner
-python-debian
-rpm-py-installer
 docker
 cryptography<3.1; python_version < '3.6'
 -r requirements-lint.txt
@@ -12,3 +6,4 @@ cryptography<3.1; python_version < '3.6'
 -r https://github.com/ansible/ansible/raw/devel/test/lib/ansible_test/_data/requirements/sanity.rstcheck.txt
 -r https://github.com/ansible/ansible/raw/devel/test/lib/ansible_test/_data/requirements/sanity.yamllint.txt
 -r https://github.com/ansible/ansible/raw/devel/test/lib/ansible_test/_data/requirements/sanity.validate-modules.txt
+molecule[docker]
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -1,0 +1,5 @@
+---
+installer_command: foreman-installer
+installer_verbose: true
+installer_no_colors: false
+installer_options: []
--- a/roles/installer/molecule/default/converge.yml
+++ b/roles/installer/molecule/default/converge.yml
@@ -1,0 +1,10 @@
+---
+- name: Converge
+  hosts: all
+  gather_facts: false
+  vars:
+    installer_scenario: foreman
+    installer_options:
+      - '--foreman-initial-admin-password changeme'
+  roles:
+    - installer
--- a/roles/installer/molecule/default/molecule.yml
+++ b/roles/installer/molecule/default/molecule.yml
@@ -1,0 +1,22 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: ${DRIVER_NAME:-podman}
+platforms:
+  - name: foreman.example.com
+    image: centos:8
+    command: /sbin/init
+    tmpfs:
+      - /run
+      - /tmp:exec,mode=777
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+provisioner:
+  name: ansible
+verifier:
+  name: ansible
+lint: |
+  set -e
+  yamllint -c ../../.yamllint .
+  ansible-lint .
--- a/roles/installer/molecule/default/prepare.yml
+++ b/roles/installer/molecule/default/prepare.yml
@@ -1,0 +1,22 @@
+---
+- name: Prepare
+  hosts: all
+  gather_facts: true
+  become: true
+  tasks:
+    - name: 'Setup Foreman latest repository'
+      yum:
+        name: https://yum.theforeman.org/releases/latest/el{{ ansible_distribution_major_version }}/x86_64/foreman-release.rpm
+        disable_gpg_check: true
+        state: present
+
+    - name: 'Setup Puppet 6 Repository'
+      yum:
+        name: https://yum.puppet.com/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
+        disable_gpg_check: true
+        state: present
+
+    - name: Install foreman-installer
+      package:
+        name: foreman-installer
+        state: present
--- a/roles/installer/molecule/default/verify.yml
+++ b/roles/installer/molecule/default/verify.yml
@@ -1,0 +1,10 @@
+---
+- name: Converge
+  hosts: all
+  gather_facts: false
+  vars:
+    installer_scenario: foreman
+    installer_verbose: true
+    installer_no_colors: false
+  roles:
+    - installer


roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: 'Run installer'
+  shell: >
+      {{ foreman_installer_command }}
+      --scenario {{ foreman_installer_scenario }}
```
mdellweg: This variable is missing from defaults.

mdellweg (commented):
For testing: Does the foreman-installer have a sort of "dry-run" or "syntax check" mode?

roles/foreman_installer/tasks/main.yml
ehelms: I was thinking of keeping this out of defaults and having the role fail if you do not specify this. Rather than defaulting to Foreman and just assuming what the user wants.

mdellweg: OK. I'd like to have a pre_task with "assert:" To provide a meaningful error message to the user.
(Not sure if you can add pre_tasks to roles)

```
pre_tasks:
  - name: Check that necessary variables are defined
    assert:
      that:
        - foreman_installer_scenario is defined
```

mdellweg: Also the README should reflect that.

ehelms:
> For testing: Does the foreman-installer have a sort of "dry-run" or "syntax check" mode?

It does have a `--noop` flag that can be passed.

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
...
+      {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
+      {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
+      {{ (foreman_installer_disable_system_checks|bool) | ternary("--disable-system-checks", "") }}
```
ehelms: @evgeni In Forklift, we had included this as defined options due to our use of them in different places. I wonder if that makes sense in this use case though? When does something get a dedicated option vs. rely on the user to specify it for their case.

roles/foreman_installer/tasks/main.yml
ehelms: Added into the role

roles/foreman_installer/tasks/main.yml
evgeni: Depends? ;)

I think having a flag/option for verbose is a good thing, colors and system checks maybe less so?

roles/foreman_installer/tasks/main.yml
evgeni: Why is `foreman` a bad default, tho? I mean, this is the foreman installer role, which kinda sounds like it's gonna install foreman by default? ;)

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: 'Run installer'
+  shell: >
```
evgeni:
```suggestion
  command: >
```

There is no need for a `shell` here

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,10 @@
...
+      {{ foreman_installer_command }}
+      --scenario {{ foreman_installer_scenario }}
+      {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
+      {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
+      {{ (foreman_installer_disable_system_checks|bool) | ternary("--disable-system-checks", "") }}
+      {{ foreman_installer_options | join(' ') }}
+  when: not foreman_installer_skip_installer
```
evgeni: With the current task list, I see no reason to expose `skip_installer` to the user?

roles/foreman_installer/tasks/main.yml
ehelms: The `foreman-installer` project does not impose a default scenario, so why should this role?

roles/foreman_installer/tasks/main.yml
ehelms: I wanted to protect the foreman-installer's back. (Friday dad jokes, you're welcome)

roles/foreman_installer/README.md
```diff
@@ -0,0 +1,28 @@
...
+
+Run the installer setting the initial admin password:
+
+```yaml
+- hosts: target-host
+  roles:
+    - role: theforeman.operations.foreman_installer
+      vars:
+        foreman_installer_options:
+          - '--foreman-initial-admin-password changeme'
```
evgeni: from a user POV, these are the things that I want to modify most often:
* admin-user
* admin-password
* primary org
* primary loc
* list of enabled plugins

I see plugins might be a bit tough as they don't have a stable naming in the cli options (partially because what users think are plugins are actually core features that can be disabled [like vmware] vs plugins [azure]).

roles/foreman_installer/tasks/main.yml
ehelms: I had meant to ask you about that. Since this could be done at the role declaration level, whether we should translate this to here from Forklift or keep it clean (sounds like keep it clean).

roles/foreman_installer/tasks/main.yml
evgeni: to provide opinionated defaults? the project does not because it wants to inform the user how to get the list of scenarios, right (or are there other reasons)? the user won't use that role for it.

roles/foreman_installer/tasks/main.yml
evgeni: in forklift, we use that to speed up things as the role there is also taking care of installing packages, and we use "skip" to run install in parallel on server and proxy, and then without skip on server and then proxy, right?

sounds like a forklift oddity that should stay there.

evgeni: and even then, this could be optimized by having the install step in a install.yml and then using `import_role` with `tasks_from: install.yml`

roles/foreman_installer/README.md
ehelms: Updated the README with some of what you describe here -- is this what you are looking for?

evgeni: What I wanted to say (but obviously didn't): I think those are the ones that a pure user would love to see as individual options/variables/flags instead of passing around an array of cli options

ehelms: Oh, ha. Yes, one for each plugin would be harder to maintain. I'm happy to add the other 4 though they don't apply to every scenario (foreman-proxy-content, capsule) so that would be my only hesitation.

roles/foreman_installer/tasks/main.yml
ehelms: I think the project does not because a user should declare what scenario they are wanting to install: foreman, katello, proxy, proxy with content. And I think, given there are many scenarios out there, a user should be forced to declare it explicitly. No surprises.

ehelms:
I got the molecule to actually run (thanks @evgeni !) but gotta track down the reasons for the errors. I am also seeing, and not sure how to fix the build errors introduced by the molecule directory:

```
install: omitting directory 'roles/foreman_installer/molecule/default'
Makefile:71: recipe for target 'build/src/roles/foreman_installer/molecule/default' failed
```

evgeni:
The build error comes from the Makefile adding the molecule folder (but not the contents) to the dependencies:
https://github.com/theforeman/foreman-operations-collection/blob/528498f9c8c93af3c346df7d9beec8e3405bedeb/Makefile#L10

Given we don't need the molecule folder in the distribution, we can probably filter it out there somehow.

evgeni:
Probably either make that last `*` a `*.*` (so it only matches files with dots) or instead of `wildcard`, move to `shell find -type f`...

@mdellweg thoughts?

mdellweg:
> Probably either make that last `*` a `*.*` (so it only matches files with dots) or instead of `wildcard`, move to `shell find -type f`...
> 
> @mdellweg thoughts?

That could work. But a role has a rather well defined directory layout. We could just be more specific what to include.

evgeni:
True, especially as now we don't include the readme of the role in the distribution

ehelms:
Thanks! I ended up trying `filter-out` in the DEPENDENCIES block and things are passing now. I landed on using molecule to do a real test of the installer inside a container. The test currently just runs the role and passes if the installer completes. I will add some additional tests for the various parameters that are hard coded... any other desires for testing?

Makefile
```diff
@@ -7,7 +7,7 @@ ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
 METADATA := galaxy.yml LICENSE README.md requirements.txt CHANGELOG.rst
 $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(eval _$(PLUGIN_TYPE) := $(filter-out %__init__.py,$(wildcard plugins/$(PLUGIN_TYPE)/*.py))))
-DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(wildcard $(ROLE)/*/*))
+DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(filter-out $(ROLE)/molecule/default, $(wildcard $(ROLE)/*/*)))
```
mdellweg: I would `filter-out` all names in molecule with `$(ROLE)/molecule/%` here.

roles/foreman_installer/tasks/main.yml
```diff
@@ -4,10 +4,19 @@
...
+  changed_when: false
+
 - name: 'Run installer' # noqa 301
   command: >
     {{ foreman_installer_command }}
     --scenario {{ foreman_installer_scenario }}
     {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
     {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
     {{ foreman_installer_options | join(' ') }}
+  when: foreman_installer_noop.rc != 0
```
mdellweg: Nice idea!!!

ehelms:
Another update round, I had to modify the `noop` method to only perform the check after the first install had at least completed.

roles/foreman_installer/README.md
evgeni: all plugin enable options are of the form `--enable-foreman-plugin-PLUGIN_NAME`, would it be in a users favor if we provide a `enable_plugins` list (empty by default), which will translate into `--enable-foreman-plugin-PLUGIN_NAME` params?

I am not sure myself, so mostly throwing out ideas. Especially as removing the plugin from that list won't disable it…

.github/workflows/main.yml
```diff
@@ -110,3 +110,21 @@ jobs:
         with:
           name: collection
           path: theforeman-operations-*.tar.gz
+
+  molecule:
+    runs-on: ubuntu-latest
+    strategy:
+      max-parallel: 4
+      matrix:
+        python-version: [3.6]
```
evgeni: does this need to be a matrix? given we only test one version?

if yes, mind using 3.7, strings and a more verbose list notation?
```suggestion
        python-version:
          - "3.7"
```

.github/workflows/main.yml
```diff
@@ -110,3 +110,21 @@ jobs:
...
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: gofrolist/molecule-action@v2
+        with:
+          molecule_working_dir: roles/foreman_installer
+        env:
+          DRIVER: docker
```
evgeni: why `ENV` instead of `molecule_options: --driver docker`?

.github/workflows/main.yml
```diff
@@ -110,3 +110,21 @@ jobs:
...
+        python-version: [3.6]
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
+        uses: actions/setup-python@v2
+        with:
+          python-version: ${{ matrix.python-version }}
+      - uses: gofrolist/molecule-action@v2
+        with:
+          molecule_working_dir: roles/foreman_installer
```
evgeni: in `forklift`, I just used a `for` loop over all roles that have a molecule folder, and no action:
https://github.com/theforeman/forklift/blob/master/.github/workflows/main.yml#L32-L38

is this action buying us anything? it's not official…

Makefile
```diff
@@ -7,7 +7,7 @@ ROLES := $(wildcard roles/*)
 PLUGIN_TYPES := $(filter-out __%,$(notdir $(wildcard plugins/*)))
 METADATA := galaxy.yml LICENSE README.md requirements.txt CHANGELOG.rst
 $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(eval _$(PLUGIN_TYPE) := $(filter-out %__init__.py,$(wildcard plugins/$(PLUGIN_TYPE)/*.py))))
-DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(wildcard $(ROLE)/*/*))
+DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(filter-out $(wildcard $(ROLE)/molecule/*), $(wildcard $(ROLE)/*/*)))
```
evgeni:
```suggestion
DEPENDENCIES := $(METADATA) $(foreach PLUGIN_TYPE,$(PLUGIN_TYPES),$(_$(PLUGIN_TYPE))) $(foreach ROLE,$(ROLES),$(filter-out $(ROLE)/molecule/%, $(wildcard $(ROLE)/*/*)))
```

I think that should also work and is shorter.

roles/foreman_installer/defaults/main.yml
```diff
@@ -0,0 +1,6 @@
+---
+foreman_installer_command: foreman-installer
+foreman_installer_verbose: true
+foreman_installer_no_colors: false
+foreman_installer_disable_system_checks: false
```
evgeni: this feature was removed?

roles/foreman_installer/molecule/default/molecule.yml
```diff
@@ -0,0 +1,37 @@
...
+    - destroy
+    - syntax
+    - create
+    - prepare
+    - converge
+    - idempotence
+    - side_effect
+    - verify
+    - cleanup
+    - destroy
```
evgeni: is this still needed, given that we have idempotency test now?

roles/foreman_installer/molecule/default/prepare.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: Prepare
+  hosts: all
+  gather_facts: true
+  become: true
+  tasks:
+    - name: 'Setup Foreman nightly repository'
```
evgeni: should we use `latest` instead of `nightly` to test a more stable installer?

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
+---
+- name: Check that necessary variables are defined
+  assert:
+    that:
+      - foreman_installer_scenario is defined
+
+- name: Check if initial install has happened
+  find:
+    paths: /etc/foreman-installer/scenarios.d
+    patterns: .installed
```
evgeni: this tests for the presence of `/etc/foreman-installer/scenarios.d/.installed`, right?

I'd have used the `stat` module and checked `result.exists`

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  when: installed.matched == 1
+
+- name: 'Run installer' # noqa 301
+  command: >
+    {{ foreman_installer_command }}
+    --scenario {{ foreman_installer_scenario }}
+    {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
+    {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
+    {{ foreman_installer_options | join(' ') }}
+  when: not foreman_installer_noop.skipped and foreman_installer_noop.rc != 0
```
evgeni:
```suggestion
  when:
    - not foreman_installer_noop.skipped 
    - foreman_installer_noop.rc != 0
```

Doesn't this mean the role never does an initial install, as this is missing the `.installed` file?

.github/workflows/main.yml
ehelms: I meant to ask you about that. The size of the matrix seemed a lot to spin up containers with full blown installs and so I started with picking a single python version to do a live test. Is that a sound strategy?

.github/workflows/main.yml
ehelms: Because:

```
ehelms@wareagle foreman_installer (add-foreman-installer)$ molecule check --driver docker
Usage: molecule check [OPTIONS]
Try 'molecule check --help' for help.

Error: no such option: --driver
ehelms@wareagle foreman_installer (add-foreman-installer)$ molecule --driver docker check
Usage: molecule [OPTIONS] COMMAND [ARGS]...
Try 'molecule --help' for help.

Error: no such option: --driver
```

.github/workflows/main.yml
evgeni: I don't think the results will vary based on Python version (we don't have any python code), so if anything, I'd test Ansible versions.

.github/workflows/main.yml
evgeni: `--driver-name`, don't take my comments literally ;)

ehelms: They must have dropped that:

```
ehelms@wareagle foreman-operations-collection (add-foreman-installer)$ molecule --driver-name docker check
Usage: molecule [OPTIONS] COMMAND [ARGS]...
Try 'molecule --help' for help.

Error: no such option: --driver-name
ehelms@wareagle foreman-operations-collection (add-foreman-installer)$ molecule check --driver-name docker
Usage: molecule check [OPTIONS]
Try 'molecule check --help' for help.

Error: no such option: --driver-name
```

roles/foreman_installer/README.md
ehelms: My leaning here is file an issue for re-evaluation but I want to start with the most basic version of this before building logic into it.

.github/workflows/main.yml
evgeni: Weird. Also if you pass that to `molecule test`?

evgeni: But if that's not working reliably, let's ot block.

Maybe should call the env var MOLECULE_DRIVER though

ehelms: `MOLECULE_` is a reserved namespace. So I opted for `DRIVER_NAME` for closer consistency.

roles/foreman_installer/README.md
evgeni: ACK

.github/workflows/main.yml
```diff
@@ -110,3 +110,32 @@ jobs:
...
+    strategy:
+      max-parallel: 4
+      matrix:
+        ansible:
+          - stable-2.9
+          - stable-2.10
+          - devel
+    steps:
+      - uses: actions/checkout@v2
+      - name: Set up Python ${{ matrix.python-version }}
```
evgeni:
```suggestion
      - name: Set up Python
```

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  when: installed.stat.exists
+
+- name: 'Run installer' # noqa 301
+  command: >
+    {{ foreman_installer_command }}
+    --scenario {{ foreman_installer_scenario }}
+    {{ (foreman_installer_verbose|bool) | ternary("-v", "") }}
+    {{ (foreman_installer_no_colors|bool) | ternary("--no-colors", "") }}
+    {{ foreman_installer_options | join(' ') }}
+  when: foreman_installer_noop.rc is not defined or foreman_installer_noop.rc != 0
```
evgeni: Should we check for `rc == 2` as in "changes have been made"?

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  assert:
+    that:
+      - foreman_installer_scenario is defined
+
+- name: Check if initial install has happened
+  stat:
+    path: /etc/foreman-installer/scenarios.d/.installed
+  register: installed
+
+- name: 'Run installer in noop mode'
```
evgeni:
```suggestion
- name: 'Check if changes are needed by running installer in noop mode'
```

roles/foreman_installer/tasks/main.yml
```diff
@@ -0,0 +1,28 @@
...
+  command: >
+    {{ foreman_installer_command }}
+    --scenario {{ foreman_installer_scenario }}
+    --noop
+  register: foreman_installer_noop
+  changed_when: false
```
evgeni:
```suggestion
  changed_when: false
  failed_when: foreman_installer_noop.rc not in [0, 2]
```

Otherwise this `command` will be marked as failed when there are actually changes to be made and the installer exited with 2. And then the play is aborted.

roles/foreman_installer/tasks/main.yml
ehelms: From the noop mode run? I think `noop` returns a `2` if changes will be made.

ehelms: Ohh, maybe you are saying just to be more explicit here. Only run this if its a 2

evgeni: Yes, that's what I was saying

roles/foreman_installer/README.md
```diff
@@ -0,0 +1,32 @@
+theforeman.operations.foreman_installer
+======================================
```
evgeni:
```suggestion
=======================================
```

roles/foreman_installer/tasks/main.yml
evgeni: Granted, given we don't allow anything but 0 or 2 in the previous step, it won't happen, but I'd still prefer to be explicit.

ehelms: However, given the current untested state of noop, and all of (https://github.com/theforeman/kafo/blob/master/README.md#exit-code) I am inclined to stick with, if `noop` does not return 0 let's run the installer. This should help prevent any cases right now where things in complex scenarios might leave users confused.

evgeni: We've written `  failed_when: foreman_installer_noop.rc not in [0, 2]` above, so the execution will be aborted if the installer exits anything but 0 or 2 and we never reach this step.

roles/installer/README.md
```diff
@@ -0,0 +1,32 @@
+theforeman.operations.installer
+=======================================
```
evgeni:
```suggestion
===============================
```

ehelms: :angry:

evgeni: Soooooorriiiiii ;)

roles/foreman_installer/tasks/main.yml
ehelms: Hrmm, you are right. You did have me do that. Fair enough.

ehelms: Updated to what I think we landed on design wise

.yamllint
```diff
@@ -0,0 +1,3 @@
+---
+rules:
+  line-length: disable
```
evgeni: :(

What would be needed to enable that check again?

roles/installer/tasks/main.yml
```diff
@@ -0,0 +1,29 @@
...
+  command: >
+    {{ installer_command }}
+    --scenario {{ installer_scenario }}
+    --noop
+  register: installer_noop
+  changed_when: false
+  failed_when: installer_noop.rc not in [0, 2]
+  when: installed.stat.exists
+
+- name: 'Run installer' # noqa 301
```
evgeni: What's 301?

evgeni (approved):
Two non blocking questions

roles/installer/tasks/main.yml
ehelms: Good reminder, that was due to pre-idempotence, now that I have that I can remove this.

.yamllint
ehelms: Adjusted to a max of 140

PR #4, opened 2020-11-26
fix CI badge in readme

evgeni:


evgeni:
so @mdellweg was like seconds faster than me

PR #5, opened 2020-11-30
added or_usermanagement

IncredibleRichie:
Hi everybody, we want to share/add this role.

### DIFF
--- a/roles/or_user_management/README.md
+++ b/roles/or_user_management/README.md
@@ -1,0 +1,148 @@
+# or_usermanagement
+
+## This role can be used to create users, user groups, LDAP-Auth and to link external and internal user groups
+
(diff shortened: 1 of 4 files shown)


ehelms:
Hi @IncredibleRichie - thanks for sharing a role! Looking over the role, this seems to use modules from [FAM](https://github.com/theforeman/foreman-ansible-modules) and provide enhancements to user management within Foreman. If I am reading that correctly, I think this would be a better candidate to go into FAM as we've started accepting [roles there](https://github.com/theforeman/foreman-ansible-modules/tree/develop/roles).

IncredibleRichie:
Oh THX! Somehow it looked right to me :-).- I will open a pr there.

PR #8, opened 2020-12-15
Update Installer Role

willtome:


roles/installer/tasks/main.yml
```diff
@@ -27,3 +32,13 @@
     {{ (installer_no_colors|bool) | ternary("--no-colors", "") }}
     {{ installer_options | join(' ') }}
   when: installer_noop.rc is not defined or installer_noop.rc == 2
+
+- name: Wait for satellite to be running
```
ehelms: This is not Satellite :) Checking Foreman is up is not a guarantee necessarily though. If backend services are not up yet,  then you may not be able to perform actions you are looking for. In puma-based Foreman installations this check should be unnecessary. Could a check of the ping API be better?

evgeni: I'd prefer not to have the check here at all (but the use the `status_info` module from the `theforeman.foreman` collection if you really want that in your workflow)

Nevermore24:
Can we at least get the installer code merged without the check? I'm dead in the water with "[Errno 2] No such file or directory: b'foreman-installer': b'foreman-installer'", "rc": 2

roles/installer/tasks/main.yml
```diff
@@ -4,6 +4,11 @@
     that:
       - installer_scenario is defined
 
+- name: Install foreman-installer
```
ehelms: `Install {{ installer_package }}` ?

roles/installer/tasks/main.yml
willtome: I think there needs to be some sort of check. If you are running a series of roles (eg. install, create org, etc) there needs to be validation that the install is complete and the application is running so that the next role doesn't go and then timeout because a service was still starting. That was the logic for a simple test against the fqdn. So I have 2 options to propose:

1) a simple check in this role to validate that the installation is done and other roles can proceed. If we use `status_info`, which I think is a great idea, do we want that dependency from this collection to the other or just move this role to the other collection? I would vote to _not_ create that dependency.
2) does a status check role go into the formeman Ansible module collection and is it the playbook authors responsibility to design a sequence like install -> status check -> other roles.

evgeni: Luckily, modern Foreman is more robust during startup, so you should not see the "installer is finished but the services aren't yet running properly" case anymore.

That said, a check using `status_info` is not hurting anything, so we can still add that (and I have no problem with the dependency or asking the user to have that in their own playbook).

evgeni (changes requested):
This needs a rebase because the `   when: installer_noop.rc is not defined or installer_noop.rc == 2` line isn't there anymore (and that PR shouldn't touch it anyways).

Other than that, I think this is ready to be merged

ehelms:
> Other than that, I think this is ready to be merged

Agreed. If you could squash this down to one commit and move it out of draft we can get this in @willtome

ehelms:
I opened https://github.com/theforeman/foreman-operations-collection/pull/27 which includes this fix and makes the updates.

PR #11, opened 2021-01-12
properly detect if an installer run is required

### LINKED ISSUE #12: installer role does not re-run the installer when needed
##### SUMMARY

The `installer` role currently checks `foreman-installer --noop` on existing installs, and only runs the real deal when that exited `2`.

This is due to three things:
1. we don't call the installer with `--detailed-exitcodes`, which means that in the case Puppet exited with [0,2], Installer will exit with 0 (https://github.com/theforeman/foreman-installer/blob/8610c990613c53c9f81b5f7f5e57918eb5e8588a/bin/foreman-installer#L11)
2. we don't pass `installer_options` to the `--noop` call, so if all that changed was the user adding an option, this won't be detected.
3. Puppet does not exit 2 when run with `--noop` (https://tickets.puppetlabs.com/browse/PUP-686)

1 and 2 are solved in #11 
3 needs a different fix, probably dropping the `--noop` logic for now, and bring it back when Kafo can gather "things would have changed" from Puppet reports

##### STEPS TO REPRODUCE

```yaml
roles:
- role: theforeman.operations.installer
  vars:
    installer_scenario: foreman
- role: theforeman.operations.installer
  vars:
    installer_scenario: foreman
    installer_options:
      - '--enable-foreman-plugin-ansible'
```

##### EXPECTED RESULTS
Ansible plugin enabled

##### ACTUAL RESULTS
Second invocation didn't trigger a real installer run

evgeni:
1. we need to pass the same installer options, as otherwise changed options
would not trigger an installer run
2. we need to enable detailed exitcodes, as otherwise installer never exits
with code 2 when there are pending changes

### DIFF
--- a/roles/installer/molecule/default/converge.yml
+++ b/roles/installer/molecule/default/converge.yml
@@ -4,7 +4,13 @@
   gather_facts: false
   vars:
     installer_scenario: foreman
-    installer_options:
-      - '--foreman-initial-admin-password changeme'
   roles:
-    - installer
+    - role: installer
+      vars:
+        installer_options:
+          - '--foreman-initial-admin-password changeme'
+    - role: installer
+      vars:
+        installer_options:
+          - '--foreman-initial-admin-password changeme'
+          - '--enable-foreman-plugin-tasks'
--- a/roles/installer/molecule/default/converge.yml
+++ b/roles/installer/molecule/default/converge.yml
@@ -6,11 +6,7 @@
     installer_scenario: foreman
   roles:
     - role: installer
-      vars:
-        installer_options:
-          - '--foreman-initial-admin-password changeme'
     - role: installer
       vars:
         installer_options:
-          - '--foreman-initial-admin-password changeme'
           - '--enable-foreman-plugin-tasks'
--- a/roles/installer/molecule/default/verify.yml
+++ b/roles/installer/molecule/default/verify.yml
@@ -8,3 +8,12 @@
     installer_no_colors: false
   roles:
     - installer
+  tasks:
+    - name: find tasks config file
+      stat:
+        path: /etc/foreman/plugins/foreman-tasks.yaml
+      register: foreman_tasks_yaml
+    - name: check tasks config file exists
+      assert:
+        that:
+          - foreman_tasks_yaml.stat.exists
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -4,23 +4,6 @@
     that:
       - installer_scenario is defined
 
-- name: Check if initial install has happened
-  stat:
-    path: /etc/foreman-installer/scenarios.d/.installed
-  register: installed
-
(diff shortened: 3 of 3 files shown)


evgeni:
draft, because it's still not sufficient

evgeni:
the problem is that `puppet apply --noop --detailed-exitcodes` still exists 0 as no changes were made: https://tickets.puppetlabs.com/browse/PUP-686

evgeni:
@ehelms @ekohl would appreciate some thoughts here. as it stands today, the role will *never* re-execute the installer, and that will lead to very unexpected results.

ekohl:
I don't know if you can. Perhaps a good start is to implement `changed_when` on the installer by adding detailed exit codes?

ehelms:
> 2\. we need to enable detailed exitcodes, as otherwise installer never exits
>     with code 2 when there are pending changes

This sounds like a bug.

ehelms:
> @ehelms @ekohl would appreciate some thoughts here. as it stands today, the role will _never_ re-execute the installer, and that will lead to very unexpected results.

Are you able to add an additional test case to show this?

ekohl:
For what it's worth, I've said it before but we should really use Puppet's reporting feature. I started with https://github.com/theforeman/kafo/pull/290 but that can reliably tell you if it would make any changes.

roles/installer/molecule/default/verify.yml
```diff
@@ -8,3 +8,12 @@
     installer_no_colors: false
   roles:
     - installer
```
evgeni, line 10: should we pass the same options as in `converge.yml` here? currently, we don't (not that the installer cares much…)

roles/installer/molecule/default/converge.yml
```diff
@@ -4,7 +4,13 @@
...
   roles:
-    - installer
+    - role: installer
+      vars:
+        installer_options:
+          - '--foreman-initial-admin-password changeme'
+    - role: installer
+      vars:
+        installer_options:
+          - '--foreman-initial-admin-password changeme'
```
evgeni: Technically, we don't need to pass the password anymore.

evgeni: more technically, we don't need to set the pass in the first call either, nobody cares for the password in the container that is thrown away after the installer run anyways.

ekohl: So why are they in here? :)

ekohl (approved):

roles/installer/molecule/default/converge.yml
evgeni: Because that was not the part I was changing, and thought we could come to a consensus before nuking :)

ehelms: I think I was anticipating re-use of this down the line where password might be used. You can nuke it.

ehelms (approved):

PR #13, opened 2021-01-15
Add role to setup Foreman repositories

### LINKED ISSUE #9: foreman repositories role
##### SUMMARY

In forklift we have a foreman_repositories (and katello_repositories) role to setup our repos, I think it'd be beneficial to offer those as part of this collection for a broader userbase to consume

##### ISSUE TYPE
 - Feature Idea

ehelms:
Starter to fix #9 and to open up a few questions as I build it out:

 1) Should we actually include staging repository setup in this collection? (its a bit of CI / dev resource)
 2) Should this role be simply named `repositories` ?
 3) Should this role handle Katello and Pulpcore repositories rather than as separate roles like we have in Forklift?

I am attempting to not just blindly copy Forklift but think about the context of this collection.

### DIFF
--- a/.yamllint
+++ b/.yamllint
@@ -1,4 +1,4 @@
 ---
 rules:
   line-length:
-    max: 140
+    max: 200
--- a/roles/foreman_repositories/README.md
+++ b/roles/foreman_repositories/README.md
@@ -1,0 +1,40 @@
+theforeman.operations.foreman_repositories
+==========================================
+
+Sets up the Foreman and plugin repositories.
+
+Role Variables
+--------------
+
+Optional:
+
+- `foreman_repositories_version`: Version of Foreman to setup repositories for (default: 2.3)
+- `foreman_repositories_katello_version`: Version of Katello to setup repositories for, set a value for this to configure Katello repositories otherwise no Katello repositories will be configured (default: null)
+
+Example Playbooks
+-----------------
+
+Setup repositories for Foreman 2.3:
+
+```
+---
+- hosts: all
+  gather_facts: true
+  vars:
+    foreman_repositories_version: "2.3"
+  roles:
+    - foreman_repositories
+```
+
+Setup repositories for Katello 3.18 and Foreman 2.3:
+
+```
+---
+- hosts: all
+  gather_facts: true
+  vars:
+    foreman_repositories_version: "2.3"
+    foreman_repositories_katello_version: "3.18"
+  roles:
+    - foreman_repositories
+```
--- a/roles/foreman_repositories/defaults/main.yml
+++ b/roles/foreman_repositories/defaults/main.yml
@@ -1,0 +1,2 @@
+---
+foreman_repositories_version: "2.3"
--- a/roles/foreman_repositories/molecule/debian/verify.yml
+++ b/roles/foreman_repositories/molecule/debian/verify.yml
@@ -1,0 +1,15 @@
+---
+- name: Verify
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name:
+      slurp:
+        path: /etc/apt/sources.list.d/deb_theforeman_org.list
+      register: repos
+
+    - name: check foreman repos exists
+      assert:
+        that:
+          - "'deb http://deb.theforeman.org buster 2.3' in repos['content'] | b64decode"
+          - "'deb http://deb.theforeman.org plugins 2.3' in repos['content'] | b64decode"
--- a/roles/foreman_repositories/tasks/debian.yml
+++ b/roles/foreman_repositories/tasks/debian.yml
@@ -1,0 +1,22 @@
+---
+- name: 'Install GPG package'
+  package:
+    name: 'gpg'
+    state: present
+    update_cache: yes
+  tags:
+    - packages
+
+- name: 'Install Foreman GPG key'
+  apt_key:
+    url: https://deb.theforeman.org/foreman.asc
+
+- name: 'Setup Foreman {{ foreman_repositories_version }} repository'
+  apt_repository:
+    repo: deb http://deb.theforeman.org {{ ansible_distribution_release }} {{ foreman_repositories_version }}
+    state: present
+
+- name: 'Setup Foreman {{ foreman_repositories_version }} plugins repository'
+  apt_repository:
+    repo: deb http://deb.theforeman.org plugins {{ foreman_repositories_version }}
+    state: present
--- a/roles/foreman_repositories/tasks/redhat.yml
+++ b/roles/foreman_repositories/tasks/redhat.yml
@@ -1,0 +1,26 @@
+---
+- name: 'Install centos-release-scl-rh' # noqa 403
+  yum:
+    name: centos-release-scl-rh
+    state: latest
+    update_cache: yes
+  tags:
+    - packages
+  when:
+    - ansible_distribution == 'CentOS'
+    - ansible_distribution_major_version == '7'
+
+- name: 'Setup Foreman {{ foreman_repositories_version }} Repository'
+  yum:
+    name: https://yum.theforeman.org/releases/{{ foreman_repositories_version }}/el{{ ansible_distribution_major_version }}/x86_64/foreman-release.rpm
+    disable_gpg_check: True
+    state: present
+  tags:
+    - packages
+
+- name: 'Setup Katello {{ foreman_repositories_katello_version }} Repository'
+  yum:
+    name: https://fedorapeople.org/groups/katello/releases/yum/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm # noqa 204
+    disable_gpg_check: True
+    state: present
+  when: foreman_repositories_katello_version is defined
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -136,6 +136,6 @@ jobs:
         run: |
           for roledir in roles/*/molecule; do
             pushd $(dirname $roledir)
-            molecule test
+            molecule test --all
             popd
           done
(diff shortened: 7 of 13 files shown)


.yamllint
```diff
@@ -1,4 +1,4 @@
 ---
 rules:
   line-length:
-    max: 140
+    max: 160
```
mdellweg: ;)

mdellweg (commented):
1. I don't know.
2. "repositories" alone can be a bit confusing given that the user is likely to work with pulp/katello
3. Separate roles would give more flexibility you can still "combine" them by inclusion in a common role.But i am unsure if it's really necessary. Maybe split it out no sooner than there is a request for it?

evgeni:
So I also had briefly looked into starting this, and also bumped `line-length` to 160…

More seriously tho:
* I don't think staging is something we should have in here (which will make forklift re-using this role a tad more complicated, but oh well)
* I think this role should not point to nightly by default (but there is no "latest" symlink in our debian repos, huh)
* I am strugling with the "repositories" name, as that might be confusing with the "repositories" role in FAM
  * why not keep `foreman_repositories`, as it's what it is: repos from theforeman.org, not epel, not puppet, etc.
* I like adding katello etc here. Makes the user experience better (plus we then could say "give me katello 3.18 repos" and it'd enable foreman 2.3 automatically

roles/foreman_repositories/molecule/debian/verify.yml
```diff
@@ -0,0 +1,25 @@
+---
+- name: Verify
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name:
+      stat:
```
evgeni: do we *need* the `stat` here? Or shall we rely on `slurp` later failing if it's not there?

roles/foreman_repositories/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+foreman_repositories_version: nightly
+foreman_repositories_katello: false
+foreman_repositories_katello_version: nightly
```
evgeni: those aren't documented

roles/foreman_repositories/README.md
```diff
@@ -0,0 +1,16 @@
...
+
+Sets up the Foreman and plugin repositories.
+
+Role Variables
+--------------
+
+Optional:
+
+- `foreman_repositories_version`: Version of Foreman to setup repositories for (default: nightly)
+- `foreman_repositories_environment`: release or staging
```
evgeni: this is gone now?

roles/foreman_repositories/README.md
```diff
@@ -0,0 +1,16 @@
...
+- `foreman_repositories_plugins`: Enable or disable setup of Foreman plugins repository (default: true)
```
evgeni: this seems gone, intentional? (I guess so)

roles/foreman_repositories/molecule/debian/prepare.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: prepare
+  hosts: all
+  gather_facts: true
+  tasks:
+    - name: install gpg
```
evgeni: kinda duplicates what's in `roles/foreman_repositories/tasks/debian.yml`?

roles/foreman_repositories/tasks/debian.yml
```diff
@@ -0,0 +1,22 @@
+---
+- name: 'Install GPG package'
+  package:
+    name: 'gpg'
+    state: present
+  tags:
+    - packages
+  when: ansible_distribution_release == "buster"
```
evgeni: I always wondered why this is limited to buster? Same should be needed on Ubuntu too?

roles/foreman_repositories/molecule/debian/prepare.yml
evgeni: do you only need the `update_cache` part for prepare in a container?

roles/foreman_repositories/tasks/debian.yml
```diff
@@ -0,0 +1,22 @@
...
+
+- name: 'Install Foreman {{ foreman_repositories_version }} repository'
+  apt_repository:
+    repo: deb http://deb.theforeman.org {{ ansible_distribution_release }} {{ foreman_repositories_version }}
+    state: present
+
+- name: 'Manage Foreman {{ foreman_repositories_version }} plugins repository'
+  apt_repository:
+    repo: deb http://deb.theforeman.org plugins {{ foreman_repositories_version }}
+    state: "{{ foreman_repositories_plugins | ternary('present', 'absent') }}"
```
evgeni: this isn't defined anymore

roles/foreman_repositories/molecule/debian/prepare.yml
ehelms: Ahh yea, that explains things. Yes, I needed the update_cache, I can drop that here and add it to the task itself.

ehelms:
@evgeni question I had before I move it out of a draft is mode adding version mapping, and which to support

 * supply a Foreman version and katello version
 * supply only a Foreman version but no Katello version
 * supply only a Katello version but no Foreman version

evgeni:
If only Katello would live in the plugins repo and we wouldn't have this discussion… ;)

I think in the ideal scenario where Katello is "jutt a plugin", the user should only define the Foreman version, and get everything else automatically.
Today, a Katello user probably doesn't care about the underlying Foreman version (or they do, but they can't change it anyways), so the most natural solution today would be "K version only".
But at the same time, can we do this magic reliably? Would we have to ship a mapping file like we do in forklift today?
I think for the first version, it's perfectly fine to make the user pass both versions, and iterate later.

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,26 @@
...
+    state: present
+  tags:
+    - packages
+
+- name: 'Setup Katello {{ katello_repositories_version }} Repository'
+  yum:
+    name: https://fedorapeople.org/groups/katello/releases/yum/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm # noqa 204
+    disable_gpg_check: True
+    state: present
+  when: foreman_repositories_katello_version
```
evgeni: `when: foreman_repositories_katello_version is defined`? then you can omit it from the defaults file?

ehelms:
I set this to 2.3 by default. I debated setting it to nothing by default, similar to installer scenario, to force users to actively set this.

roles/foreman_repositories/README.md
```diff
@@ -0,0 +1,40 @@
...
+
+Sets up the Foreman and plugin repositories.
+
+Role Variables
+--------------
+
+Optional:
+
+- `foreman_repositories_version`: Version of Foreman to setup repositories for (default: 2.3)
+- `foreman_repositories_katello_version`: Version of Katello to setup repositories for (default: )
```
evgeni: This should somehow tell the user that if they don't set a version, no Katello repos are configured.

roles/foreman_repositories/molecule/debian/prepare.yml
evgeni: didn't you want to make this go away? :)

evgeni (changes requested):

roles/foreman_repositories/README.md
ehelms: Check the wording now

evgeni:
I don't exactly understand why CI fails (looks like it can't open a socket to fedorapeople, but why?!)
And I'd still remove the `  when: ansible_distribution_release == "buster"` on the GPG install task, but otherwise this looks good to me!

roles/foreman_repositories/tasks/debian.yml
```diff
@@ -0,0 +1,23 @@
...
+    update_cache: yes
+  tags:
+    - packages
+  when: ansible_distribution_release == "buster"
+
+- name: 'Install Foreman GPG key'
+  apt_key:
+    url: https://deb.theforeman.org/foreman.asc
+
+- name: 'Install Foreman {{ foreman_repositories_version }} repository'
```
evgeni:
```suggestion
- name: 'Configure Foreman {{ foreman_repositories_version }} repository'
```

roles/foreman_repositories/tasks/debian.yml
```diff
@@ -0,0 +1,23 @@
...
+  apt_repository:
+    repo: deb http://deb.theforeman.org {{ ansible_distribution_release }} {{ foreman_repositories_version }}
+    state: present
+
+- name: 'Manage Foreman {{ foreman_repositories_version }} plugins repository'
```
evgeni:
```suggestion
- name: 'Configure Foreman {{ foreman_repositories_version }} plugins repository'
```

roles/foreman_repositories/tasks/debian.yml
evgeni: (or "Setup" as in the RPM tasks)

ehelms:
This https://github.com/theforeman/foreman-operations-collection/pull/14 has the same code and no issues. Very odd.

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,26 @@
...
+
+- name: 'Setup Foreman {{ foreman_repositories_version }} Repository'
+  yum:
+    name: https://yum.theforeman.org/releases/{{ foreman_repositories_version }}/el{{ ansible_distribution_major_version }}/x86_64/foreman-release.rpm
+    disable_gpg_check: True
+    state: present
+  tags:
+    - packages
+
+- name: 'Setup Katello {{ katello_repositories_version }} Repository'
```
evgeni:
```suggestion
- name: 'Setup Katello {{ foreman_repositories_katello_version }} Repository'
```

evgeni (approved):
nitpick in one of the messages, ACK otherwise

PR #14, opened 2021-01-20
Add role to generate certificate bundles

ehelms:
Looking for thoughts on design and working on the tests. Current test failures are slightly contradictory.

### DIFF
--- a/roles/foreman_proxy_certs_generate/molecule/default/prepare.yml
+++ b/roles/foreman_proxy_certs_generate/molecule/default/prepare.yml
@@ -1,0 +1,19 @@
+---
+- name: Prepare
+  hosts: all
+  gather_facts: true
+  become: true
+  vars:
+    foreman_repositories_version: nightly
+    foreman_repositories_katello_version: nightly
+  roles:
+    - foreman_repositories
+    - puppet_repositories
+  post_tasks:
+    - include_role:
+        name: installer
+      vars:
+        installer_scenario: katello
+        installer_package: foreman-installer-katello
+        installer_options:
+          - '--disable-system-checks'
--- a/roles/foreman_proxy_certs_generate/tasks/main.yml
+++ b/roles/foreman_proxy_certs_generate/tasks/main.yml
@@ -1,0 +1,9 @@
+---
+- name: 'Generate certificates for a Foreman Proxy'
+  command: |
+    {{ foreman_proxy_certs_generate_command }}
+    --foreman-proxy-fqdn {{ foreman_proxy_certs_generate_fqdn }}
+    --certs-tar {{ foreman_proxy_certs_generate_tarball }}
+    {{ foreman_proxy_certs_generate_options | join(' ') }}
+  args:
+    creates: "{{ foreman_proxy_certs_generate_tarball }}"
--- a/roles/foreman_repositories/tasks/redhat.yml
+++ b/roles/foreman_repositories/tasks/redhat.yml
@@ -24,3 +24,32 @@
     disable_gpg_check: true
     state: present
   when: foreman_repositories_katello_version is defined
+
+- when:
+    - foreman_repositories_katello_version is defined
+    - ansible_distribution_major_version == '8'
+  block:
+    - name: Install dnf-config-manager
+      package:
+        name: dnf-plugins-core
+        state: present
+
+    - name: Get pki-core module info
+      command: dnf module list --enabled
+      register: enabled_modules
+      changed_when: false
+
+    - name: Enable pki-core module
+      command: dnf module enable pki-core -y
+      when: "'pki-core' not in enabled_modules.stdout"
+
+    - name: List disabled repositories
+      command: dnf repolist --disabled
+      register: disabled_repos
+      changed_when: false
+
+    - name: Enable powertools
+      command: dnf config-manager --set-enabled powertools
+      when:
+        - ansible_distribution == 'CentOS'
+        - "'powertools' in disabled_repos.stdout"
--- a/roles/foreman_proxy_certs_generate/README.md
+++ b/roles/foreman_proxy_certs_generate/README.md
@@ -1,0 +1,31 @@
+theforeman.operations.foreman_proxy_certs_generate
+==================================================
+
+Generates a certificate bundle for Katello scenario
+
+Role Variables
+--------------
+
+Required:
+
+- `foreman_proxy_certs_generate_fqdn`: Target FQDN to generate a certificate bundle for
+
+Optional:
+
+- `foreman_proxy_certs_generate_tarball`: Full path of the tarball that should be generated. Defaults to /root/<foreman_proxy_certs_generate_fqdn>.tar.gz
+- `foreman_proxy_certs_generate_options`: Array of options to pass to the installer
+
+Example Playbooks
+-----------------
+
+Generate a certificate bundle for the hostname proxy.example.com:
+
+```
+---
+- hosts: all
+  gather_facts: false
+  vars:
+    foreman_proxy_certs_generate_fqdn: proxy.example.com
+  roles:
+    - role: foreman_proxy_certs_generate
+```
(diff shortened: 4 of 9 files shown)


roles/certificate_bundle/tasks/main.yml
```diff
@@ -0,0 +1,5 @@
+---
+- name: 'Generate Certs'
+  command: foreman-proxy-certs-generate --foreman-proxy-fqdn {{ certificate_bundle_fqdn }} --certs-tar /root/{{ certificate_bundle_fqdn }}.tar.gz {{ certificate_bundle_options }} # noqa 204
```
evgeni: I think I'd prefer to allow the user to configure where the tarball lands.

so something like

```
certificate_bundle_tar: "/root/{{ certificate_bundle_fqdn }}.tar.gz"
```

in defaults and then using it here?

roles/certificate_bundle/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+certificate_bundle_options: ''
```
evgeni: in `installer`, this is an array, should we be consistent? we should :)

roles/certificate_bundle/README.md
```diff
@@ -0,0 +1,18 @@
...
+Role Variables
+--------------
+
+Required:
+
+- `certificate_bundle_fqdn`: Target FQDN to generate a certificate bundle for
+
+Optional:
+
+- `certificate_bundle_options`: Array of options to pass to the installer
```
evgeni:
```
- `certificate_bundle_command`: command to run, can be used by derivative projects to specify a branded command
```

to keep in line with the installer?

evgeni:
Is `certificate_bundle` a good name? the users are used to `<project>-certs-generate` CLI command, so maybe `proxy_certs_generate` is cleaner? also shows that this is only applicable to proxy bundles (but jaja, we don't call it proxy in some projects)

ehelms:
Ehhh you are probably right. I was trying to make it a thing rather than sounding like an action but folks do know it as certs_generate (Which keeps it agnostic?)

evgeni:
I think my main issue with `certs_generate` is that people might think this is also doing certs for the main server, which it certainly does not?

@wbclark what's your opinion?

wbclark:
> I think my main issue with `certs_generate` is that people might think this is also doing certs for the main server, which it certainly does not?
> 
> @wbclark what's your opinion?

I think naming things consistently with the names already used for the underlying tools is really essential... anything else will promote confusion and uncertainty.

I agree that the existing names are not ideal, but imo, changing the names here and not there just makes the situation even worse.

roles/foreman_proxy_certs_generate/molecule/default/prepare.yml
```diff
@@ -0,0 +1,23 @@
...
+    foreman_repositories_version: nightly
+    foreman_repositories_katello_version: nightly
+  roles:
+    - foreman_repositories
+    - puppet_repositories
+  post_tasks:
+    - name: Install katello
+      package:
+        name: foreman-installer-katello
+        state: present
```
evgeni: you can now pass the package name to the installer role ;)

roles/foreman_proxy_certs_generate/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Set tarball path
+  set_fact:
+    foreman_proxy_certs_generate_tarball: /root/{{ foreman_proxy_certs_generate_fqdn }}.tar.gz
+  when: foreman_proxy_certs_generate_tarball is undefined
```
evgeni: I think you can put that into defaults, then you'd save a `set_fact` and that ugly `when` XD

roles/foreman_proxy_certs_generate/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
...
+
+- name: 'Generate certificates for a Foreman Proxy'
+  command: |
+    foreman-proxy-certs-generate
+    --foreman-proxy-fqdn {{ foreman_proxy_certs_generate_fqdn }}
+    --certs-tar {{ foreman_proxy_certs_generate_tarball }}
+    {{ foreman_proxy_certs_generate_options | join(' ') }}
+  args:
+    creates: /root/{{ foreman_proxy_certs_generate_fqdn }}.tar.gz
```
evgeni:
```suggestion
    creates: "{{ foreman_proxy_certs_generate_tarball }}"
```

roles/foreman_proxy_certs_generate/tasks/main.yml
```diff
@@ -0,0 +1,14 @@
+---
+- name: Set tarball path
+  set_fact:
+    foreman_proxy_certs_generate_tarball: /root/{{ foreman_proxy_certs_generate_fqdn }}.tar.gz
+  when: foreman_proxy_certs_generate_tarball is undefined
+
+- name: 'Generate certificates for a Foreman Proxy'
+  command: |
+    foreman-proxy-certs-generate
```
evgeni: should the name of the command be overrideable?

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -24,3 +24,32 @@
...
+        state: present
+
+    - name: Get pki-core module info
+      command: dnf module list --enabled
+      register: enabled_modules
+      changed_when: false
+
+    - name: Enable pki-core module
+      command: dnf module enable pki-core -y
+      when: "'pki-core' not in enabled_modules.stdout"
```
evgeni, line 44: doesn't the installer do that for us these days?

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -24,3 +24,32 @@
...
+
+    - name: List disabled repositories
+      command: dnf repolist --disabled
+      register: disabled_repos
+      changed_when: false
+
+    - name: Enable powertools
+      command: dnf config-manager --set-enabled powertools
```
evgeni, line 52: can't this be done with `yum_repository` module or worst case `ini_file`?

roles/foreman_repositories/tasks/redhat.yml
ehelms: I thought it should but it fails without this.

roles/foreman_repositories/tasks/redhat.yml
ehelms: I dont think it can with `yum_repository` as you have to define the baseurl or mirrorlist yourself. Unless we want to do that rather than just enabling what is there? `ini_file` seems like a weird choice when I can use dnf directly to handle things.

roles/foreman_proxy_certs_generate/tasks/main.yml
ehelms: Depends who the stakeholder is :P But I can fix that. I was just so excited to get this working.

roles/foreman_repositories/tasks/redhat.yml
evgeni: ah, indeed, that's a rather stupid limitation of the module

fair enough.

roles/foreman_repositories/tasks/redhat.yml
evgeni: weird, but nothing we have to fix today.

evgeni (approved):

roles/foreman_proxy_certs_generate/tasks/main.yml
mdellweg: :+1: The first thing i learned with ansible is "avoid set_fact". You can never use `--start-at-task` if you have it.

evgeni: Let me translate that for you: "never use set_fact OR start-at-task" ;-)

PR #15, opened 2021-01-22
Add role to setup Puppet repositories

ehelms:


roles/puppet_repositories/README.md
```diff
@@ -0,0 +1,36 @@
...
+=========================================
+
+Sets up Puppet repositories
+
+Role Variables
+--------------
+
+Optional:
+
+- `puppet_repositories_version`: Version of Foreman to setup repositories for (default: 2.3)
```
evgeni: Uhm, no, I don't think this will setup Foreman 2.3 repos ;)

roles/puppet_repositories/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: 'Include puppet {{ puppet_repositories_version }} for {{ ansible_os_family }}'
+  include_tasks: 'puppetlabs-{{ puppet_repositories_version }}-{{ ansible_os_family|lower }}.yml'
```
evgeni: there is no 7 file but you refer to 7 in the examples? :)

roles/puppet_repositories/README.md
ehelms: It might if you wish hard enough.

roles/puppet_repositories/tasks/puppetlabs-6-redhat.yml
```diff
@@ -0,0 +1,6 @@
+---
+- name: 'Setup Puppet 6 Repository'
+  yum:
+    name: https://yum.puppet.com/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
```
evgeni: I wonder if we need a versioned file these days, or if that can be:
```suggestion
    name: https://yum.puppet.com/puppet{{ puppet_repositories_version }}-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
```

roles/puppet_repositories/tasks/puppetlabs-6-debian.yml
```diff
@@ -0,0 +1,15 @@
+---
```
evgeni: there are `puppetX-relase-${DIST}.deb` files on https://apt.puppet.com/ and I wonder if using them in a similar way as we do with RPM would be a good idea.

Saves us the `apt_key` call as the deb already has the key in the right format, and also the installation onf `gpg`.

roles/puppet_repositories/tasks/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+- name: 'Include {{ ansible_os_family }}'
+  include_tasks: '{{ ansible_os_family|lower }}.yml'
+  when: puppet_repositories_version is defined
```
evgeni: given there is a default, it will always be default

.github/workflows/main.yml
```diff
@@ -134,8 +134,9 @@ jobs:
         env:
           DRIVER_NAME: docker
         run: |
-          for roledir in roles/*/molecule; do
-            pushd $(dirname $roledir)
+          roles=$(git diff develop --stat | grep roles | cut -d "/" -f 2 | uniq)
```
evgeni: This means it won't run after a merge (as the diff is empty in that case) and do the wrong thing for PRs against non-develop branches (you could use [GITHUB_BASE_REF](https://docs.github.com/en/actions/reference/environment-variables) to avoid that).

ehelms: Do we run the tests after a merge?

evgeni: yes:
https://github.com/theforeman/foreman-operations-collection/blob/9dce0978cb7ebfe2f49e835c31f9de0da0fc8097/.github/workflows/main.yml#L3

(the `push` part is what would run on merge)

ehelms: Hrmm, well I am gonna drop this for now. I am also thinking about how we may use other roles in our tests for other roles. We should probably ensure those dont break.

evgeni (approved):

evgeni:
Ignoring (unrelated) CI error and merging.

PR #21, opened 2021-02-19
Add role to upgrade postgresql on EL8

ehelms:


ehelms:
@evgeni I think CI here is failing hard due to some changes that I, from looking in FAM, think you have overcome. When I tried to mimic some of those requirements and Makefile updates here I kept making a mess of things. Could you maybe give things a once over for any changes you know we need to make?

evgeni:
@ehelms the `lint` issue should be fixed in https://github.com/theforeman/foreman-operations-collection/pull/23, but I can't wrap my head around how to fix `molecule`.

ehelms:
Includes https://github.com/theforeman/foreman-operations-collection/pull/23 with some additional fixes I read about. Let me know thoughts on how I tackled it. There is some repetition that can be cleaned up later.

ehelms:
I debated if I should drop the lint job given molecule does linting.

.github/workflows/main.yml
```diff
@@ -119,7 +121,6 @@ jobs:
         ansible:
           - stable-2.9
           - stable-2.10
-          - devel
```
evgeni: I guess this is temporally?

roles/postgresql_upgrade/README.md
```diff
@@ -0,0 +1,19 @@
+theforeman.operations.postgresql_upgrade
+========================================
+
+Upgrades Postgresql for a Foreman install.
```
evgeni:
```suggestion
Upgrades PostgreSQL for a Foreman install.
```

roles/postgresql_upgrade/molecule/redhat/prepare.yml
```diff
@@ -0,0 +1,25 @@
...
+      yum:
+        name: https://yum.theforeman.org/releases/latest/el{{ ansible_distribution_major_version }}/x86_64/foreman-release.rpm
+        disable_gpg_check: true
+        state: present
+
+    - name: 'Setup Puppet 6 Repository'
+      yum:
+        name: https://yum.puppet.com/puppet6-release-el-{{ ansible_distribution_major_version }}.noarch.rpm
+        disable_gpg_check: true
+        state: present
```
evgeni: we have roles for that now ;)

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
+---
+- name: Get version of postgres data directory
+  slurp:
+    path: /var/lib/pgsql/data/PG_VERSION
+  register: pg_version
+
+- when:
+    - ansible_distribution_major_version == '8'
```
evgeni: This should also check for this being *RHEL* 8, not Debian 8 (even if we don't support Debian 8 anymore)

Plus I'd put this check as the first task, `/var/lib/pgsql` won't exist on EL7 installs with SCL'ed PostgreSQL anymore.

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,58 @@
...
+      set_fact:
+        collate: "{{ postgres_output.stdout.split('|')[3] | trim }}"
+        ctype: "{{ postgres_output.stdout.split('|')[4] | trim }}"
+
+    - name: Stop postgresql
+      service:
+        name: postgresql
+        state: stopped
+
+    - name: Update postgresq.conf data directory
```
evgeni, line 38:
```suggestion
    - name: Update postgresql.conf data directory
```

Update to what? The task essentially deletes the config line.

roles/postgresql_upgrade/README.md
evgeni: I think this also should point out that this is only for EL8 systems with module-based PSQL, which will update 10 to 12, not any arbitrary versions.

.github/workflows/main.yml
ehelms: Aye.

.github/workflows/main.yml
```diff
@@ -102,7 +102,9 @@ jobs:
           restore-keys: |
             ${{ runner.os }}-pip-
       - name: Install dependencies
-        run: pip install --upgrade -r requirements-lint.txt
+        run: |
+          pip install ansible
```
ekohl: Why is it needed to install ansible separate?

ehelms:
12 runs later it is green again

.github/workflows/main.yml
evgeni: ansible-lint doesn't depend on it (anymore), but putting it into lint.txt would pull it in other places where we want the git checkout of ansible instead.

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -23,4 +23,5 @@
     name: https://fedorapeople.org/groups/katello/releases/yum/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm
     disable_gpg_check: true
     state: present
+  retries: 3
```
evgeni: it feels wrong to put in a workaround for our CI into the role itself, not that I'd have a better solution :(

roles/postgresql_upgrade/README.md
```diff
@@ -0,0 +1,19 @@
+theforeman.operations.postgresql_upgrade
+========================================
+
+Upgrades PostgreSQL for a Foreman install on EL8 only.
```
evgeni:
```suggestion
Upgrade Foreman's PostgreSQL from 10 to 12 on EL8.
```

roles/postgresql_upgrade/README.md
```diff
@@ -0,0 +1,19 @@
...
+
+Role Variables
+--------------
+
```
evgeni:
```suggestion
```

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
+---
+- name: Get version of postgres data directory
```
evgeni: I thought more about the following:

```yaml
- name: Ensure running on EL8
  assert:
    that:
      - ansible_os_family == "RedHat"
      - ansible_distribution_major_version == '8'
    fail_msg: "This role is only supported on EL8"

- name: Get version from the PostgreSQL data directory
  slurp:
    path: /var/lib/pgsql/data/PG_VERSION
  register: pg_version

- name: Ensure PostgreSQL version is 10
  assert:
    that:
      - "'10' in pg_version['content']|b64decode"
    fail_msg: "This role only supports upgrades from PostgreSQL 10"

[ rest of the tasks without the block ]
```

But wouldn't block on my personal preferences ;)

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+    - ansible_distribution_major_version == '8'
+    - "'10' in pg_version['content']|b64decode"
+  block:
+    - name: Start postgresql
+      service:
+        name: postgresql
+        state: started
+
+    - name: Collect postgres data
+      shell: psql -lt | grep -E "^\s+postgres"
```
evgeni:
```suggestion
      shell: psql --list --tuples-only | grep -E "^\s+postgres"
```

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+        state: present
+
+    - name: Update postgresql packages  # noqa package-latest
+      dnf:
+        name:
+          - postgresql-server
+          - postgresql-upgrade
+        state: latest
+
+    - name: Run pg_upgrade
```
evgeni:
```suggestion
    - name: Run postgresql-setup --upgrade
```

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+        name: postgresql
+        state: stopped
+
+    - name: Remove data directory entry from postgresql.conf
+      lineinfile:
+        path: /var/lib/pgsql/data/postgresql.conf
+        regexp: '^data_directory ='
+        line: ''
+
+    - name: Reset postgresql module
```
ekohl: This doesn't look like a reset but rather an enable.

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,62 @@
...
+      dnf:
+        name: '@postgresql:12'
+        state: present
+
+    - name: Update postgresql packages  # noqa package-latest
```
ekohl, line 49: This is odd. For me a `dnf module install postgresql:12` also upgraded all packages to the new version. Then it was sufficient for me to ensure `postgresql-update` was installed.

evgeni:
I left a few nitpicky comments. I think the only thing worth debating is the addition of "retries" to the repositories role, which seems wrong to me, but gets the CI to a working setup…

ekohl (commented):
Should the source and target version be variables? This seems like a good practice to me. My programming professor would fail us if we had a number other than 0, 1 or 2 in the code. They all had to be constants. It's something I still feel quite strongly about and it pains me to see `12` in multiple places.

roles/foreman_repositories/tasks/redhat.yml
ehelms: I was hoping you would, cause re-running the job 12 times yesterday felt bad and this felt just enough less bad to be worth it.

One solution is to move Katello repos off of fedorapeople :P

roles/postgresql_upgrade/tasks/main.yml
ehelms: The initial assert makes sense to me. The last assert will break idempotency I think?

evgeni: True.

evgeni: True.

.github/workflows/main.yml
ekohl: The correct way to deal with is to call `pip install --upgrade -r requirements-lint.txt ansible`, not 2 separate calls to pip. This allows the pip's dependency resolution to work correctly.

ehelms:
Rebased on the latest repository changes

evgeni:
Seems this push here:
```
 @ehelms ehelms force-pushed the ehelms:postgres-upgrade branch from 8ee7e6e to dd33362 5 hours ago 
```

has overwritten a lot of changes we discussed and you did prior to the Katello repo fix :(

ehelms:
I think I recovered all the updates I lost through the comments.

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+      - ansible_distribution_major_version == '8'
+    fail_msg: "This role is only supported on EL8"
+
+- name: Get version from the PostgreSQL data directory
+  slurp:
+    path: /var/lib/pgsql/data/PG_VERSION
+  register: pg_version
+  when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
```
evgeni:
```suggestion
```

The `assert` above will already have bailed out if that's not the case

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -0,0 +1,69 @@
...
+
+- when:
+    - ansible_os_family == "RedHat"
+    - ansible_distribution_major_version == '8'
```
evgeni:
```suggestion
```

The `assert` above will already have bailed out if that's not the case

ehelms: This doesn't hurt though :)

roles/postgresql_upgrade/tasks/main.yml
ehelms: This doesn't hurt though :)

roles/postgresql_upgrade/molecule/redhat/prepare.yml
```diff
@@ -0,0 +1,10 @@
+---
+- name: Prepare
+  hosts: all
+  gather_facts: true
+  roles:
+    - puppet_repositories
+    - foreman_repositories
+    - role: installer
+      vars:
+        installer_scenario: foreman
```
evgeni: should this explicitly point at a foreman version that came with psql 10, as we plan to shift the default to 12 soon?

roles/postgresql_upgrade/tasks/main.yml
evgeni: Sure, but my eyes/brain ;)

roles/postgresql_upgrade/molecule/redhat/prepare.yml
ehelms: Good point, I'll point it at 2.3.

roles/postgresql_upgrade/tasks/main.yml
ehelms: Better to be explicit and safe? Just never look at this code again ;)

evgeni (approved):
I'd *prefer* if the `when`s would be trimmed, but won't block on it.

PR #22, opened 2021-02-19
Support ansible-lint 5

ehelms:


evgeni (approved):
the molecule error is not related to this change

PR #23, opened 2021-03-05
CI fixes

evgeni:


mdellweg:
if ansible-core is installed, it returns `ansible [core 2.11.0b1]` on `ansible --version` and ansible-lint is not able to cope with this.

Failing here:
https://github.com/ansible-community/ansible-lint/blob/master/src/ansiblelint/config.py#L97

requirements-dev.txt
```diff
@@ -5,4 +5,4 @@ cryptography<3.1; python_version < '3.6'
 -r requirements.txt
 pylint==2.6.0; python_version >= '3.6'
 -r https://github.com/ansible/ansible/raw/devel/test/lib/ansible_test/_data/requirements/sanity.validate-modules.txt
-molecule[docker]
+molecule[docker]>=3.2.4
```
ehelms: Seems to not like this.

ehelms:
I guess we need to add yamllint as an explicit requirement?

requirements-dev.txt
evgeni: "python 2.7", yeah

ehelms: I don't think that version supports Python 2.7. Seems like it was last supported here https://pypi.org/project/molecule/3.0.2.1/. So gotta keep it the way it was or drop Py 2.7

ehelms (approved):
Katello repos failures is as Katello repo failures does.

PR #24, opened 2021-03-09
Correct README headers + use syntax highlighting

ekohl:


ehelms:
ACK to the syntax highlighting. As I have understood things, the current header method I have chosen is the standard way in Ansible roles to do it. That how it is in FAM and I feel like that is what @evgeni told me to do.

evgeni:
Yeah, the headers are what `ansible-galaxy role init` would do for us (see https://raw.githubusercontent.com/ansible/ansible/devel/lib/ansible/galaxy/data/default/role/README.md). I am not married to it tho, given it should produce the same output.

ehelms:
Doing syntax highlighting here -- https://github.com/theforeman/foreman-operations-collection/pull/36 and choosing to keep the headers the same convention we have across collections for now.

PR #25, opened 2021-03-09
Fixes #20: Support upgrading puppet major versions

### LINKED ISSUE #20: support upgrading major puppet versions
##### SUMMARY

Puppets `puppetX-release` RPMs (and probably Debs, untested) conflict with all other non-X versions (6 conflicts with 5, 7 with 6 and 6), but don't obsolete those. So just installing the new one while the old is installed doesn't work (yum transaction error). We need to remove the old version first.

In `forklift` we just have erased 5 when we installed 6: https://github.com/theforeman/forklift/blob/517d8ddcb77f6fb97525cb4efd35db5212c301fb/roles/puppet_repositories/tasks/puppetlabs-6-redhat.yml#L2-L7

To make it more generic here, I propose to define a list of supported versions (e.g `puppet_repositories_supported_versions: [6,7]`) and then have a task that erases all `puppetX-release` packages where X is in `{{ puppet_repositories_supported_versions | difference([puppet_repositories_version]) }}`.

##### ISSUE TYPE
 - Feature Idea

ehelms:
Based on design from https://github.com/theforeman/foreman-operations-collection/issues/20 -- I have intentionally kept it to Puppet 5 and 6 right now while Puppet 7 support is being developed for Foreman.

### DIFF
--- a/roles/puppet_repositories/molecule/redhat/verify.yml
+++ b/roles/puppet_repositories/molecule/redhat/verify.yml
@@ -3,7 +3,12 @@
   hosts: all
   gather_facts: false
   tasks:
-    - name: Ensure puppet release RPM is installed
+    - name: Ensure puppet5-release RPM is absent
+      package:
+        name: puppet5-release
+        state: absent
+
+    - name: Ensure puppet6-release RPM is installed
       package:
         name: puppet6-release
         state: present
--- a/roles/puppet_repositories/vars/main.yml
+++ b/roles/puppet_repositories/vars/main.yml
@@ -1,0 +1,3 @@
+puppet_repositories_supported_versions:
+  - "5"
+  - "6"
--- a/roles/puppet_repositories/molecule/debian/prepare.yml
+++ b/roles/puppet_repositories/molecule/debian/prepare.yml
@@ -1,8 +1,8 @@
 ---
-- name: Converge
+- name: Prepare
   hosts: all
   gather_facts: true
   vars:
-    puppet_repositories_version: 6
+    puppet_repositories_version: 5
   roles:
     - puppet_repositories
--- a/roles/puppet_repositories/molecule/debian/verify.yml
+++ b/roles/puppet_repositories/molecule/debian/verify.yml
@@ -3,12 +3,12 @@
   hosts: all
   gather_facts: false
   tasks:
-    - name:
-      slurp:
-        path: /etc/apt/sources.list.d/puppet6.list
-      register: repos
+    - name: Ensure puppet5-release package is absent
+      package:
+        name: puppet5-release
+        state: absent
 
-    - name: check puppet repo exists
-      assert:
-        that:
-          - "'deb http://apt.puppetlabs.com buster puppet6' in repos['content'] | b64decode"
+    - name: Ensure puppet6-release package is present
+      package:
+        name: puppet6-release
+        state: present
--- a/roles/puppet_repositories/molecule/redhat/converge.yml
+++ b/roles/puppet_repositories/molecule/redhat/converge.yml
@@ -2,5 +2,7 @@
 - name: Converge
   hosts: all
   gather_facts: true
+  vars:
+    puppet_repositories_version: 6
   roles:
     - puppet_repositories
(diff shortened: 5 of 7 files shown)


evgeni:
I like!

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -18,9 +18,28 @@
   tags:
     - packages
 
-- name: 'Setup Katello {{ foreman_repositories_katello_version }} Repository'
-  yum:
-    name: https://fedorapeople.org/groups/katello/releases/yum/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm
-    disable_gpg_check: true
-    state: present
-  when: foreman_repositories_katello_version is defined
+- when: foreman_repositories_katello_version is defined
```
ehelms: @evgeni I know this is rather ugly. The first run of it worked. Trying again as well. The problem I face is that I am blocked making progress on this PR and others without this approach given CI won't test what we need it to test. Thoughts?

ehelms:
Rebased on the recent develop such that this should pass reliably now.

roles/puppet_repositories/molecule/redhat/verify.yml
```diff
@@ -3,7 +3,12 @@
   hosts: all
   gather_facts: false
   tasks:
-    - name: Ensure puppet release RPM is installed
+    - name: Ensure puppet6-release RPM is installed
       package:
         name: puppet6-release
         state: present
+
+    - name: Ensure puppet5-release RPM is absent
```
evgeni, line 6: Should this be the first test? You have it the other way round in debian/verify.yml (nitpick, obviously)

roles/puppet_repositories/tasks/debian.yml
```diff
@@ -5,6 +5,12 @@
     state: present
     update_cache: true
 
+- name: Remove any existing Puppet release packages
```
evgeni:
```suggestion
- name: Remove any existing Puppet release packages for versions other than {{ puppet_repositories_version }}
```

Otherwise users will freak out XD

roles/puppet_repositories/tasks/redhat.yml
```diff
@@ -1,4 +1,10 @@
 ---
+- name: Remove any existing Puppet release RPMs that are not for the version
```
evgeni:
```suggestion
- name: Remove any existing Puppet release packages for versions other than {{ puppet_repositories_version }}
```

To sync with debian

evgeni: technically this task can go to `main.yml` as it is correct for both targets

roles/puppet_repositories/vars/main.yml
```diff
@@ -0,0 +1 @@
+puppet_repositories_supported_versions: [5, 6]
```
evgeni:
```suggestion
puppet_repositories_supported_versions:
  - 5
  - 6
```

I prefer long lists, but that's not what my brain jumped at me.

You use integers here, but nothing forbids the user to pass a string as the desired version, and then the `difference` above will not do what we want:

```console
% ansible -m debug -a 'msg={{ [1,2] | difference(["2"]) }}' localhost
localhost | SUCCESS => {
    "msg": [
        1,
        2
    ]
}
```

So I guess we need to hard-convert things to strings at some point. But IMHO this can go as an improvement in a later PR.

evgeni (approved):

roles/puppet_repositories/tasks/debian.yml
```diff
@@ -5,6 +5,12 @@
...
+  package:
+    name: "puppet{{ item }}-release"
+    state: absent
+  loop: "{{ puppet_repositories_supported_versions | difference([puppet_repositories_version]) }}"
```
evgeni:
```suggestion
  loop: "{{ puppet_repositories_supported_versions | map('string') | difference([puppet_repositories_version | string]) }}"
```

this should fix the issue I was talking about

evgeni (approved):

PR #26, opened 2021-03-23
Fixes #32172 - use Katello repos on yum.theforeman.org

evgeni:


evgeni:
`-latest` is not available yet, but soon

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -20,7 +20,7 @@
 
 - name: 'Setup Katello {{ foreman_repositories_katello_version }} Repository'
   yum:
-    name: https://fedorapeople.org/groups/katello/releases/yum/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm
+    name: https://yum.theforeman.org/katello/{{ foreman_repositories_katello_version }}/katello/el{{ ansible_distribution_major_version }}/x86_64/katello-repos-latest.rpm
```
ehelms, line 23: I suppose we should ensure 3.17 exists there to so that this role represents supported versions ?

evgeni: Yeah, there is a 3.17.3 in flight, which will autopublish there too.

ehelms (approved):

PR #27, opened 2021-03-25
Add package install to installer role

ehelms:


ehelms:
Replaces https://github.com/theforeman/foreman-operations-collection/pull/8 with a few minor fix ups

evgeni (approved):
should the readme mention the default value too?

ehelms:
> should the readme mention the default value too?

I debated that and left it be since we weren't mentioning others either right now.

PR #29, opened 2021-03-26
Fix #28: Scope role variables with foreman_

### LINKED ISSUE #28: Scope role variable naming to be prefixed with `foreman_`

ehelms:


ehelms:
I'd like to get a release out with the latest round of updates but I think we should make this change before hand.

roles/puppet_repositories/README.md
```diff
@@ -1,4 +1,4 @@
-theforeman.operations.puppet_repositories
+theforeman.operations.foreman_puppet_repositories
```
evgeni: nope

ehelms:
`foreman_` scoping seems natural, but I also wondered if it should be `foreman_operations` scoping.

evgeni:
> `foreman_` scoping seems natural, but I also wondered if it should be `foreman_operations` scoping.

I don't mind either way. Given those roles probably won't have a "common" set of variables (contrary to FAM), we could even use role-specific prefixes and nothing would break.

ehelms:
> we could even use role-specific prefixes and nothing would break.

This is what we have today and I suppose we'd just need to change a few of the roles names? e.g. `installer` makes sense in the collection naming `theforeman.operations.installer` but not as variable names `installer_`. I lean towards just having consistency of naming with prefixes.

ehelms:
@mdellweg @wbclark Your thoughts would be appreciated.

wbclark:
I too could go either way with it. For example, `foreman_installer_options` feels very natural, while `foreman_operations_installer_options` is a little cumbersome.

I'm trying to think of specific examples or user stories where having disambiguation in variable naming across the two collections has a clear benefit.

mdellweg:
I just stumbled on the `foreman_installer_scenario` variable in the `postgres_upgrade` role. Is this a shared variable, or is it already a manifestation of a name clash? `foreman_postgres_upgrade_installer_scenario` might be the "correct" scoping, but its rather unwieldy. If we go with `foreman_` as a prefix (and i think that is what i prefer) would it be `foreman_foreman_installer_scenario`?
I'm already confused by that example

ehelms:
> I just stumbled on the `foreman_installer_scenario` variable in the `postgres_upgrade` role. Is this a shared variable, or is it already a manifestation of a name clash?

This is defining the value for the `foreman_installer_scenario` variable coming from the `installer` role. It's not a clash or an overlap of naming. The snippet from the molecule testing shows it being set and passed to the installer role:

```
    - role: installer
      vars:
        foreman_installer_scenario: foreman
```

mdellweg:
> > I just stumbled on the `foreman_installer_scenario` variable in the `postgres_upgrade` role. Is this a shared variable, or is it already a manifestation of a name clash?
> 
> This is defining the value for the `foreman_installer_scenario` variable coming from the `installer` role. It's not a clash or an overlap of naming.

A closer look helps. Thanks.

ehelms:
The consensus seems to be stick with what is in this PR and match the standards we've established in FAM. That works for me and allows me to proceed. If I get a review that would be appreciated :)

mdellweg (approved):

evgeni (approved):

PR #30, opened 2021-04-06
add ansible 2.11 to test matrix

evgeni:
also stop skipping sanity on python 3.8, it was fixed


mdellweg (approved):

wbclark (approved):
I can see that the intended tests ran, including sanity on python 3.8

By the way, what is the reason to test ansible devel on older python versions where we do not test any stable ansible release?

PR #32, opened 2021-04-07
Add enablement of Ruby 2.7 module for foreman-installer 2.5+

ehelms:
@evgeni Thoughts here? Feels ugly to add yet another variable to represent version, and felt weird to include `foreman_repositories_version` in the `installer` role. I thought about just introducing a shared variable `foreman_version` and `katello_version` to use across any roles.

### DIFF
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -4,6 +4,22 @@
     that:
       - foreman_installer_scenario is defined
 
+- when:
+    - ansible_distribution_major_version == '8'
+    - ansible_os_family == 'RedHat'
+  block:
+    - name: Get foreman-installer version
+      dnf:
+        list: foreman-installer
+      register: package_info
+
+    - name: Enable ruby:2.7 module
+      dnf:
+        name: '@ruby:2.7'
+        state: present
+      when:
+        - (package_info.results | map(attribute='version') | max) is version("2.5", ">=")
+
 - name: "Install {{ foreman_installer_package }}"
   package:
     name: "{{ foreman_installer_package }}"
--- a/roles/foreman_repositories/README.md
+++ b/roles/foreman_repositories/README.md
@@ -8,33 +8,33 @@ Role Variables
 
 Optional:
 
-- `foreman_repositories_version`: Version of Foreman to setup repositories for (default: 2.3)
+- `foreman_repositories_version`: Version of Foreman to setup repositories for (default: 2.4)
 - `foreman_repositories_katello_version`: Version of Katello to setup repositories for, set a value for this to configure Katello repositories otherwise no Katello repositories will be configured (default: null)
 
 Example Playbooks
 -----------------
 
-Setup repositories for Foreman 2.3:
+Setup repositories for Foreman 2.4:
 
 ```
 ---
 - hosts: all
   gather_facts: true
   vars:
-    foreman_repositories_version: "2.3"
+    foreman_repositories_version: "2.4"
   roles:
     - foreman_repositories
 ```
 
-Setup repositories for Katello 3.18 and Foreman 2.3:
+Setup repositories for Katello 4.0 and Foreman 2.4:
 
 ```
 ---
 - hosts: all
   gather_facts: true
   vars:
-    foreman_repositories_version: "2.3"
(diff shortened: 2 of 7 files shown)


evgeni:
I'm still mad we can't properly express that with RPM metadata and have to fall back to hacks like this.

Using a shared version variable across installer and repos sounds like the cleanest approach here.

Unless there is a way to find the version of the installable (but yet not installed) foreman-installer package? https://docs.ansible.com/ansible/latest/collections/ansible/builtin/package_facts_module.html only reports installed packages...

evgeni:
```
# ansible -m dnf -a list=epel-release localhost
localhost | SUCCESS => {
    "changed": false,
    "msg": "",
    "results": [
        {
            "arch": "noarch",
            "epoch": "0",
            "name": "epel-release",
            "nevra": "0:epel-release-8-8.el8.noarch",
            "release": "8.el8",
            "repo": "extras",
            "version": "8",
            "yumstate": "available"
        }
    ]
}
```

I guess that could be used?

evgeni:
```
- hosts: localhost
  tasks:
    - name: find epel-release
      dnf:
        list: epel-release
      register: versions
    - name: output
      debug:
        msg: "latest epel-release is {{ versions.results | map(attribute='version') | max }}"
```

;)

ehelms:
Thanks for the find. I updated to use that `list` method for looking up what version is available. I also included two commits to adjust default version tested and to use the repositories roles.

roles/installer/tasks/main.yml
```diff
@@ -4,6 +4,22 @@
...
+      package:
+        list: foreman-installer
+      register: package_info
+
+    - name: Enable ruby:2.7 module
+      dnf:
+        name: '@ruby:2.7'
+        state: present
+      when:
+        - package_info.results[0]['version'] == 'nightly' or package_info.results[0]['version'] is version('2.5', '>=')
```
evgeni: The RPM version will never be "nightly" ;)

roles/installer/defaults/main.yml
```diff
@@ -4,3 +4,4 @@ foreman_installer_package: foreman-installer
 foreman_installer_verbose: true
 foreman_installer_no_colors: false
 foreman_installer_options: []
+foreman_installer_version: "2.3"
```
evgeni: this can go now?

roles/installer/tasks/main.yml
```diff
@@ -4,6 +4,22 @@
     that:
       - foreman_installer_scenario is defined
 
+- when:
+    - ansible_distribution_major_version == '8'
+    - ansible_os_family == 'RedHat'
+  block:
+    - name: Get foreman-installer version
+      package:
+        list: foreman-installer
```
evgeni, line 13: `list` is not documented for `package`, just for `yum` and `dnf`, and while this works, I'd prefer to stick to `dnf` here.

evgeni:
molecule failures unrelated, should be fixed by https://github.com/theforeman/foreman-operations-collection/pull/33

roles/installer/tasks/main.yml
evgeni: and my example above used map/max to find the newest available version, just to be super sure for setups that have multiple repos enabled…

evgeni:
You need to set `gather_facts: true` in https://github.com/theforeman/foreman-operations-collection/blob/94403a1d99f2616968d03c9bcb839605b39fb73b/roles/installer/molecule/default/verify.yml#L4 otherwise the run will fail.

evgeni:
Actually, no, why do we include the role in verify?! That's nonsense. Converge already tested idempotency.

evgeni:
rebasiing on https://github.com/theforeman/foreman-operations-collection/pull/34 should do it.

evgeni:
Yay, 🍏!

Is there a way we could add a test for the 2.5+ behavior?

ehelms:
> Is there a way we could add a test for the 2.5+ behavior?

Depends. I can add a test of things where I use nightly, or we revisit the test once 2.5 RC is available.

evgeni:
Testing on nightly until a better target is available sounds fine to me.

evgeni (approved):

PR #33, opened 2021-04-07
don't use nightly for tests

evgeni:


ehelms (approved):
Thanks for remembering we can do that now

PR #34, opened 2021-04-08
don't run installer role when verifying the installer

evgeni:
instead check that the foreman package is installed

### DIFF
--- a/roles/installer/molecule/default/verify.yml
+++ b/roles/installer/molecule/default/verify.yml
@@ -1,14 +1,22 @@
 ---
-- name: Converge
+- name: Verify
   hosts: all
   gather_facts: false
-  vars:
-    foreman_installer_scenario: foreman
-    foreman_installer_verbose: true
-    foreman_installer_no_colors: false
-  roles:
-    - installer
   tasks:
+    - name: Foreman is installed
+      package:
+        name: foreman
+        state: present
+
+    - name: find installer flag file
+      stat:
+        path: /etc/foreman-installer/scenarios.d/.installed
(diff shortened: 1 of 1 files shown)


wbclark (approved):
I guess it's safe to assume we won't have any change in packaging which causes the Foreman package to get installed before the installer is run. I always find that counter intuitive, coming from downstream where most packages get installed early as dependencies of the Satellite RPM.

A brief comment here explaining the logic in case we come back to it much later might be nice

ehelms (approved):

roles/installer/molecule/default/verify.yml
```diff
@@ -1,14 +1,13 @@
...
-    foreman_installer_scenario: foreman
-    foreman_installer_verbose: true
-    foreman_installer_no_colors: false
-  roles:
-    - installer
   tasks:
+    - name: Foreman is installed
+      package:
+        name: foreman
+        state: present
```
ehelms, line 9: Maybe we should verify some artifact of the installer is present? Like https://github.com/theforeman/foreman-installer/blob/51d002636f178c0084483cb01f8aed7395a14900/hooks/boot/01-kafo-hook-extensions.rb#L11

ehelms: Or that we can access the web UI.

evgeni: I thought about the `.installed` file too, can add.

UI seems a bit overkill (and potentially slow in the container)

evgeni: added

PR #38, opened 2021-04-29
Added environment for foreman-installer

Thulium-Drake:
When using this on my systems Ansible was unable to run the ```foreman-installer``` command because in it's environment ```LANG``` and ```LC_ALL``` were set to ```C``` instead of a valid locale.

This commit adds a default locale for both variables, you can override it by setting ```foreman_installer_locale```.


Thulium-Drake:
My other PR #43 supersedes this one, closing it as such

PR #43, opened 2021-08-07
[WIP] Various fixes

Thulium-Drake:
This PR is a collection of fixes for issues I ran into while deploying Foreman 2.4/2.5 on CentOS8 and Rocky8.

* Fixed environment for foreman-installer invocation
* Improved handling of PowerTools repo on EL8
* Added installation of Ansible 2.9 repo on EL8 for Foreman 2.5+
* ~Improved handling of PostgreSQL initialization/upgrade on EL8~
* Added preliminary support for Rocky8

Feedback is appreciated :-)

### DIFF
--- a/roles/ansible_repositories/molecule/redhat/molecule.yml
+++ b/roles/ansible_repositories/molecule/redhat/molecule.yml
@@ -1,0 +1,30 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: ${DRIVER_NAME:-podman}
+platforms:
+  - name: centos8
+    image: quay.io/centos/centos:stream8
+    command: /sbin/init
+    tmpfs:
+      - /run
+      - /tmp:exec,mode=777
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+  - name: centos7
+    image: centos:7
+    command: /sbin/init
+    tmpfs:
+      - /run
+      - /tmp:exec,mode=777
+    volumes:
+      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+provisioner:
+  name: ansible
+verifier:
+  name: ansible
+lint: |
+  set -e
+  yamllint -c ../../.yamllint .
+  ansible-lint .
--- a/roles/ansible_repositories/molecule/redhat/verify.yml
+++ b/roles/ansible_repositories/molecule/redhat/verify.yml
@@ -1,0 +1,9 @@
+---
+- name: Verify
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Ensure centos-release-ansible-29 RPM is installed
+      package:
+        name: centos-release-ansible-29
+        state: present
--- a/roles/ansible_repositories/tasks/redhat.yml
+++ b/roles/ansible_repositories/tasks/redhat.yml
@@ -1,0 +1,10 @@
+---
+- name: 'Install centos-release-ansible29 on EL8'  # noqa package-latest
+  dnf:
+    name: centos-release-ansible-29
+    state: latest
+    update_cache: true
+  tags:
+    - packages
+  when:
+    - ansible_distribution_major_version == '8'
--- a/roles/foreman_repositories/tasks/redhat.yml
+++ b/roles/foreman_repositories/tasks/redhat.yml
@@ -1,5 +1,5 @@
 ---
-- name: 'Install centos-release-scl-rh'  # noqa package-latest
+- name: 'Install centos-release-scl-rh on CentOS7'  # noqa package-latest
   yum:
     name: centos-release-scl-rh
     state: latest
@@ -51,5 +51,5 @@
     - name: Enable powertools
       command: dnf config-manager --set-enabled powertools
       when:
-        - ansible_distribution == 'CentOS'
         - "'powertools' in disabled_repos.stdout"
+        - ansible_distribution != 'RedHat'
--- a/roles/installer/molecule/default/prepare.yml
+++ b/roles/installer/molecule/default/prepare.yml
@@ -6,3 +6,9 @@
   roles:
     - puppet_repositories
     - foreman_repositories
+  tasks:
+    - name: Ensure langpacks on EL8
+      dnf:
+        name: glibc-langpack-en
+        state: present
+      when: ansible_facts['os_family'] == 'RedHat'
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -36,3 +36,6 @@
   register: foreman_installer_run
   changed_when: foreman_installer_run.rc == 2
   failed_when: foreman_installer_run.rc not in [0, 2]
+  environment:
+    LC_ALL: "{{ foreman_installer_locale }}"
+    LANG: "{{ foreman_installer_locale }}"
--- a/roles/ansible_repositories/README.md
+++ b/roles/ansible_repositories/README.md
@@ -1,0 +1,22 @@
+theforeman.operations.ansible_repositories
+=========================================
+
+Sets up Ansible repositories
+
+Role Variables
+--------------
+
+None
+
+Example Playbooks
+-----------------
+
+Setup repositories for default Ansible 2.9 for use by installer:
+
+```yaml
+---
+- hosts: all
+  gather_facts: true
+  roles:
+    - ansible_repositories
+```
--- a/roles/ansible_repositories/molecule/redhat/collections.yml
+++ b/roles/ansible_repositories/molecule/redhat/collections.yml
@@ -1,0 +1,5 @@
+collections:
+  - name: community.docker
+    version: ">=1.2.0,<2"
+  - name: community.general
+    version: ">=2,<3"
--- a/roles/ansible_repositories/molecule/redhat/converge.yml
+++ b/roles/ansible_repositories/molecule/redhat/converge.yml
@@ -1,0 +1,6 @@
+---
+- name: Converge
+  hosts: all
+  gather_facts: true
+  roles:
+    - ansible_repositories
--- a/roles/ansible_repositories/molecule/redhat/prepare.yml
+++ b/roles/ansible_repositories/molecule/redhat/prepare.yml
@@ -1,0 +1,6 @@
+---
+- name: Prepare
+  hosts: all
+  gather_facts: true
+  roles:
+    - ansible_repositories
--- a/roles/ansible_repositories/tasks/main.yml
+++ b/roles/ansible_repositories/tasks/main.yml
@@ -1,0 +1,3 @@
+---
+- name: 'Include {{ ansible_os_family }}'
+  include_tasks: '{{ ansible_os_family|lower }}.yml'
--- a/roles/foreman_proxy_certs_generate/molecule/default/prepare.yml
+++ b/roles/foreman_proxy_certs_generate/molecule/default/prepare.yml
@@ -10,6 +10,11 @@
     - foreman_repositories
     - puppet_repositories
   post_tasks:
+    - name: Ensure langpacks on EL8
+      dnf:
+        name: glibc-langpack-en
+        state: present
+      when: ansible_facts['os_family'] == 'RedHat'
     - include_role:
         name: installer
       vars:
--- a/roles/installer/README.md
+++ b/roles/installer/README.md
@@ -17,6 +17,7 @@ Optional:
 - `foreman_installer_verbose`: Enables verbose output mode in the installer
 - `foreman_installer_no_colors`: Disables color output from the installer
 - `foreman_installer_command`: Installer command to run, can be used by derivative projects to specify a branded command
+- `foreman_installer_locale`: Locale to run the installer with, this must not be ```C```, defaults to ```en_US.UTF-8```
 
 Example Playbooks
 -----------------
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -4,3 +4,4 @@ foreman_installer_package: foreman-installer
 foreman_installer_verbose: true
 foreman_installer_no_colors: false
 foreman_installer_options: []
+foreman_installer_locale: 'en_US.UTF-8'


roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +72,14 @@
       register: disabled_repos
       changed_when: false
 
-    - name: Enable powertools
-      command: dnf config-manager --set-enabled powertools
+    - name: Enable PowerTools on CentOS
+      command: dnf config-manager --set-enabled PowerTools
```
ehelms: I am still seeing this as `powertools` in my environments.

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -10,6 +10,30 @@
...
+- name: 'Install centos-release-ansible29 on CentOS8'
+  yum:
+    name: centos-release-ansible-29
+    state: latest
+    update_cache: true
+  tags:
+    - packages
+  when:
+    - ansible_distribution == 'CentOS'
+    - ansible_distribution_major_version == '8'
```
ehelms: I think a variable here to let users decide if they want the Ansible repository enabled would be good. As there may be users of this collection that do not need this repository present.

roles/postgresql_upgrade/tasks/main.yml
```diff
@@ -6,40 +6,53 @@
...
     path: /var/lib/pgsql/data/PG_VERSION
   register: pg_version
+  when: pg_present.stat.exists
 
-- when:
-    - "'10' in pg_version['content']|b64decode"
+- when: >
+    "pg_version is defined and
+    '10' in pg_version['content']|b64decode or
+    not pg_present.stat.exists"
```
ehelms: Can you expand on why these extra conditions are needed?

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Hmm, interesting, I ran into errors as it was in CamelCase on CentOS8.3 system. Anyway, I'll stick it in a rescue block then :-)

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Right! I'm exclusively working with Katello in my setups, forgot about that ;-)

roles/postgresql_upgrade/tasks/main.yml
Thulium-Drake: When running this role after the upgrade already has been done, you can run into all kinds of issues (and errors stopping the play).

All these conditions are to make sure that the upgrade will only run if the ```pg_version``` has been set and not the correct version OR when the PG data directory does not exist.

I'll also update the logic to check for the wanted version instead of the assumed lower version, makes more sense

ehelms: The tests for this role should be running an idempotency test (https://github.com/theforeman/foreman-operations-collection/tree/develop/roles/postgresql_upgrade/molecule/redhat) to verify that after it runs once code is not executed again to do any upgrading. What do you think is missing or what scenario is missing in order to ensure the tests catch this?

Thulium-Drake: You're right, it does exactly that :-) (I think I got a bit side-tracked working on other things and mixed up a few thoughts). The new feature I added is that the role will now also install postgres if it is unable to find an existing database directory.

I'm currently using the collection to install fresh Foreman/Katello servers from scratch, the tasks list I use is here: https://github.com/Thulium-Drake/ansible-role-foreman/blob/master/tasks/install_foreman.yml

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -10,6 +10,32 @@
...
+    - ansible_repository_enabled
```
ehelms: This will need to be added to defaults and the README. Per convention, we should name it `foreman_repositories_ansible_repository-enabled`. BUT, given that its a stand-alone set of repositories, I would prefer we just introduce a simple role `ansible_repositories` and handle all of this there. That greatly simplifies scoping and would be inline with things like https://github.com/theforeman/foreman-operations-collection/tree/develop/roles/puppet_repositories

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Moved to new role

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Done!

Thulium-Drake:
@ehelms Can you have another look at this? :)

Thulium-Drake:
@evgeni Can you have a look at this? <3

roles/ansible_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,24 @@
+---
+- name: 'Install centos-release-ansible29 on CentOS8'
+  yum:
```
ehelms: Since this is targeted at EL8 OS should this use `dnf` ?

roles/ansible_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,24 @@
...
+    state: latest
+    update_cache: true
+  tags:
+    - packages
+  when:
+    - ansible_distribution == 'CentOS'
+    - ansible_distribution_major_version == '8'
+
+- name: 'Install centos-release-ansible29 on Rocky8'
+  yum:
```
ehelms: Since this is targeted at EL8 OS should this use `dnf` ?

ehelms:
Minor nitpicks. @evgeni could you take a glance as well

roles/ansible_repositories/tasks/redhat.yml
Thulium-Drake: Fair enough, I've updated it! :)

roles/ansible_repositories/tasks/redhat.yml
Thulium-Drake: Fair enough, I've updated it! :)

ehelms:
We did some CI updates in the past few days, any failures you see should be actual failures with the PR now versus infra failures.

roles/ansible_repositories/tasks/redhat.yml
```diff
@@ -0,0 +1,24 @@
...
+    name:
+      - http://mirror.centos.org/centos/8/extras/x86_64/os/Packages/centos-release-configmanagement-1-1.el8.noarch.rpm
+      - http://mirror.centos.org/centos/8/extras/x86_64/os/Packages/centos-release-ansible-29-1-2.el8.noarch.rpm
+    disable_gpg_check: true
+    state: present
+  tags:
+    - packages
+  when:
+    - ansible_distribution == 'Rocky'
+    - ansible_distribution_major_version == '8'
```
evgeni: according to https://www.reddit.com/r/RockyLinux/comments/pny121/rocky_linux_addons_rockyrelease_centos_sig_content/, you should just be able to get the centos release bits from rocky extras?

evgeni: and the same for Alma: https://www.reddit.com/r/AlmaLinux/comments/pg4ev7/announcing_centos_sig_repository_availability_in/

Thulium-Drake: Oh nice! I didn't know and it wasn't in place when I first wrote this :)

I have updated the task to 'just' install centos-release-ansible-29 now :)

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +48,12 @@
...
 
-    - name: Enable powertools
-      command: dnf config-manager --set-enabled powertools
+    - name: Enable PowerTools
+      block:
+        - name: Enable PowerTools
+          command: dnf config-manager --set-enabled powertools
+      rescue:
+        - name: Enable PowerTools
+          command: dnf config-manager --set-enabled PowerTools
```
evgeni: Who's using "PowerTools" these days? C8, C8Stream, Rocky and Alma all have "powertools"

Thulium-Drake: You're right, when I wrote this, I ran into this and made this as a fallback.

I've removed it :)

roles/ansible_repositories/molecule/redhat/verify.yml
```diff
@@ -0,0 +1,9 @@
+---
+- name: Verify
+  hosts: all
+  gather_facts: false
+  tasks:
+    - name: Ensure centos-release-ansible29 RPM is installed
+      package:
+        name: centos-release-ansible29
```
evgeni:
```suggestion
        name: centos-release-ansible-29
```

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +48,9 @@
       register: disabled_repos
       changed_when: false
 
-    - name: Enable powertools
-      command: dnf config-manager --set-enabled powertools
+    - name: Enable PowerTools
+      block:
```
evgeni: the block isn't needed anymore

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +48,9 @@
...
+        - name: Enable PowerTools
+          command: dnf config-manager --set-enabled powertools
       when:
-        - ansible_distribution == 'CentOS'
-        - "'powertools' in disabled_repos.stdout"
+        - "'PowerTools' in disabled_repos.stdout"
```
evgeni: I think you want to check `powertools` not `PowerTools` here ;)

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +48,9 @@
...
       changed_when: false
 
-    - name: Enable powertools
-      command: dnf config-manager --set-enabled powertools
+    - name: Enable PowerTools
+      block:
+        - name: Enable PowerTools
+          command: dnf config-manager --set-enabled powertools
       when:
-        - ansible_distribution == 'CentOS'
```
evgeni, line 54: You're dropping this, to make it run on Rocky and Alma too, right?

I think then we should guard on `ansible_distribution != 'RedHat'`, as there is no powertools there.

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Yeah, you're right :) both are on the same line, but the repo name is ```powertools```

roles/foreman_repositories/tasks/redhat.yml
Thulium-Drake: Ah, right, I didn't knew that :) added!

Thulium-Drake:
@evgeni @ehelms Ping! :)

Thulium-Drake:
Rebased

roles/postgresql_upgrade/tasks/main.yml
evgeni: Why tho? The installer takes care of all things PostgreSQL (besides the upgrade, which this role is for).

Thulium-Drake: Should the procedure then be:

1. Install Foreman
2. Upgrade PostgreSQL

In the current playbook I first ensure that PostgreSQL is present and upgraded before running the installer. And in that case, this role goes bust when there is no PostgreSQL installed on the system:

```
TASK [Upgrade postgres on EL8] ***************************************************************************************

TASK [theforeman.operations.postgresql_upgrade : Ensure running on EL8] **********************************************
ok: [deploy.lbhr.htm.lan] => changed=false 
  msg: All assertions passed

TASK [theforeman.operations.postgresql_upgrade : Get version from the PostgreSQL data directory] *********************
fatal: [deploy.lbhr.htm.lan]: FAILED! => changed=false 
  msg: 'file not found: /var/lib/pgsql/data/PG_VERSION'
[WARNING]: Failure using method (v2_runner_on_failed) in callback plugin
(<ansible.plugins.callback.ara_default.CallbackModule object at 0x7fcc409c92e0>): 'NoneType' object is not
subscriptable
```

And with the changes I made to the role it checks if the server is already present, if not, it sets it up

evgeni: This role is only needed when one has a Foreman on EL8 which was installed with PostgreSQL 10 and wants to upgrade to one that expects 12.

By default modern Foreman always uses 12, so this role becomes a noop (we have been including it in our CI unconditionally, without any issues).

But yes, the role is not expected to be executed on empty systems.

Thulium-Drake: Right, that makes things a lot clearer! Thank you!  I'll revert the changes I made with regards to the postgresl_upgrade role :)

evgeni: Is there anything in the readme of the role that led you on the wrong path and we should improve instead?

evgeni: Is there anything in the readme of the role that led you on the wrong path and we should improve instead?

Thulium-Drake:
@evgeni I have updated the file (which removed the convo it seems). But to answer your question, I read the description of the role as "this will make sure that the EL8 system is running the correct version of psql". So when assembling a playbook that would install Foreman using the roles I assumed I had to run this first before attempting the install

roles/postgresql_upgrade/tasks/main.yml
Thulium-Drake: Ah, sorry, I'm blind, I posted an answer in the main thead :sweat_smile: 
> @evgeni I have updated the file (which removed the convo it seems). But to answer your question, I read the description of the role as "this will make sure that the EL8 system is running the correct version of psql". So when assembling a playbook that would install Foreman using the roles I assumed I had to run this first before attempting the install

roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -48,8 +48,8 @@
       register: disabled_repos
       changed_when: false
 
-    - name: Enable powertools
+    - name: Enable PowerTools
```
evgeni:
```suggestion
    - name: Enable powertools
```

just to limit the diff noise :)

roles/ansible_repositories/molecule/redhat/molecule.yml
```diff
@@ -0,0 +1,30 @@
+---
+dependency:
+  name: galaxy
+driver:
+  name: ${DRIVER_NAME:-podman}
+platforms:
+  - name: centos8
+    image: centos:8
```
evgeni:
```suggestion
    image: quay.io/centos/centos:stream8
```

roles/installer/tasks/main.yml
```diff
@@ -36,3 +36,6 @@
   register: foreman_installer_run
   changed_when: foreman_installer_run.rc == 2
   failed_when: foreman_installer_run.rc not in [0, 2]
+  environment:
+    LC_ALL: "{{ foreman_installer_locale | default('en_US.UTF-8') }}"
+    LANG: "{{ foreman_installer_locale | default('en_US.UTF-8') }}"
```
evgeni: somehow, this freaks out the installer:
```
The LANG environment variable should not be set to C
```
probably, because there is no `en_US.UTF-8` in the default stream8 container, and it falls back to C

evgeni: `dnf install glibc-langpack-en` should do it (only on EL8!)

Thulium-Drake: Interesting, when I check the langpacks on Rocky, ```glibc-all-langpacks``` is already present :-)

evgeni: in a container? I would guess a normal VM has those installed, but a minimal container, as we use it for tests does not.

evgeni: I think we can just add it as a prep step in the molecule playbook, as it's not something needed on "normal" installs

Thulium-Drake: No, a 'normal' ISO installation, makes sense that it's missing in minimal containers. I'll add it and update the PR

Thulium-Drake: I think it's best to add it as a task to the ```installer``` role, do you agree?

evgeni: I am torn (as I don't exactly know how the langpack packages work), like what happens if the user has "all" installed and we make the role install "en"?

Thulium-Drake: Maybe we should not make a change there then outside of the CI/CD scope, most (if not all) OS installers already take care of this (I haven't seen any issues with this so far, and I have deployed _a lot_ of Foremen on CentOS7, CentOS8 and Rocky8. Same goes for Satellite).

roles/installer/tasks/main.yml
```diff
@@ -36,3 +36,6 @@
   register: foreman_installer_run
   changed_when: foreman_installer_run.rc == 2
   failed_when: foreman_installer_run.rc not in [0, 2]
+  environment:
+    LC_ALL: "{{ foreman_installer_locale | default('en_US.UTF-8') }}"
```
evgeni: I think the default should be defined in `roles/installer/defaults/main.yml` and documented in `README.md`

Thulium-Drake: Done!

Thulium-Drake: Done!

evgeni (approved):

roles/installer/molecule/default/prepare.yml
```diff
@@ -8,3 +8,9 @@
   roles:
     - puppet_repositories
     - foreman_repositories
+  tasks:
+    - name: Ensure langpacks on EL8
```
evgeni, line 10: can you add the same hack to the `foreman_proxy_certs_generate` role please?

Thulium-Drake: Done! :rocket:

evgeni:
Thanks! (I've re-written history to be three commits only ;-))

Thulium-Drake:
Thank you very much! :-D

PR #46, opened 2021-10-29
only check if powertools is disabled

evgeni:
otherwise the "if powertools in stdout" condition matches
powertools-source, which we leave disabled


evgeni:
CI is broken, you prolly need to pull some of the recent fixed we did in FAM…

ehelms (approved):
Fixing CI in another PR, I am good with this change

PR #47, opened 2021-10-29
Various CI fixes adding new Ansible versions and from FAM repository

ehelms:
I was gonna break these out trying to parser changes from FAM but lumped it together, lets see if it fixes CI @evgeni

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -14,34 +14,26 @@ jobs:
       fail-fast: false
       matrix:
         python:
-          - "2.7"
-          - "3.6"
-          - "3.7"
           - "3.8"
         ansible:
           - stable-2.9
           - stable-2.10
           - stable-2.11
+          - stable-2.12
           - devel
-        exclude:
-          - python: "2.7"
-            ansible: "stable-2.9"
-          - python: "2.7"
-            ansible: "stable-2.10"
+        include:
           - python: "2.7"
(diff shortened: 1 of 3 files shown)


evgeni:
The Ansible changes look good, no idea why Molecule is failing in the installer tho

.github/workflows/main.yml
```diff
@@ -127,15 +119,16 @@ jobs:
...
       - uses: actions/checkout@v2
       - name: Set up Python
-        uses: actions/setup-python@v2.2.2
+        uses: actions/setup-python@v2
         with:
-          python-version: "3.7"
+          python-version: "3.8"
       - name: Install Ansible
-        run: pip install --upgrade git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
+        run: pip install --upgrade -r ansible
```
evgeni: this looks wrong. `-r` means requirements file, but ansible is none.

and without `-r` this would install *some* ansible, not the one we want in this matrix

evgeni:
the ansible team is currently trying to fix the 5.0.0 problem that 5.0.0 is confusing installations on py 2.7 and 3.5, see https://github.com/ansible-community/community-topics/issues/54

ehelms:
The molecule failure is due to https://github.com/theforeman/puppet-foreman/pull/1009

evgeni (approved):

PR #55, opened 2021-12-16
Add variable for python interpreter

bagasse:
RHEL8 don't provides `python` executable, only `python2` or `python3`

This permit to override the default `python` command to the correct local value.


mdellweg (approved):
I have the impression, i've seen this before.

evgeni (approved):
If only we'd have a dedicated, central Makefile for such things…

evgeni:
@bagasse there is also https://GitHub.com/pulp/squeezer ;)

Issue #54, opened 2021-12-15
Port foreman-prepare-realm command to ansible role


bagasse:
##### SUMMARY
Today, to integrate foreman with FreeIPA as a realm provider, we have to run an interactive command `foreman-prepare-realm` (that ask the password of FreeIPA user with admin rights)  to make some changes in FreeIPA that allow that integration.

It would be nice if we can use an ansible role to do that instead to be able to automate this kind of setup. I've already the ansible code for that (and also for FreeIPA DNS provider), but it introduce a dependency to [freeipa.ansible_freeipa](https://github.com/freeipa/ansible-freeipa) collection.

##### ISSUE TYPE
 - Feature Idea

evgeni:
Thanks @bagasse!

To sum what I've said on IRC (and what I got suggested in #ansible-community):
* the operations collection seems like a good place for your role, as it helps with operating a foreman instance ;)
* the dependency on `freeipa.ansible_freeipa` seems a bit "fat" for just that one role. in most "classical" package management situations, you'd declare this as an "optional" dependency, but Ansible seems not to support this
* so we probably should not declare this as a hard dependency, but document it and make the role raise a nice error if the dependency is not satisfied
* `community.general` (since 4.0) has a `a_module` test and a `collection_version` lookup, which would help with that, but we don't want a dependency on `community.general` (especially not when we want to also be able to upload this to Automation Hub as `redhat.satellite_operations`), but we still can take the code as inspiration

PR #57, opened 2021-12-16
Add prepare_realm role

bagasse:
As discussed in #54 add `prepare_realm` role.

**TODO:**
* [ ] Fail nicely if the FreeIPA collection dependency is not present.
* [ ] Testing (need FreeIPA instance and register foreman server as client)


evgeni:
@bagasse would you mind rebasing this once? so that CI has a chance ;)

bagasse:
Hello @evgeni, yes, but I've some work to do on the tests part, but I've not so much time right now to get it done...

PR #58, opened 2021-12-23
Add a role to install rh_cloud_rhc_worker

ehelms:
@shimshtein here is a starter draft roles

### DIFF
--- a/roles/cloud_connector/README.md
+++ b/roles/cloud_connector/README.md
@@ -1,0 +1,29 @@
+theforeman.operations.cloud_connector
+=====================================
+
+Install and configure Red Hat Cloud Connector
+
+Role Variables
+--------------
+
+Required:
+
+- `foreman_cloud_connector_url`: The URL of the Foreman server.
+- `foreman_cloud_connector_user`: The username cloud connector will use to talk to Foreman API.
+- `foreman_cloud_connector_password`: The password cloud connector will use to talk to Foreman API.
+
+Example Playbooks
+-----------------
+
+Run the installer setting the initial admin password:
+Configure Cloud Connector:
+
+```yaml
+- hosts: target-host
+  roles:
+    - role: theforeman.operations.cloud_connector
+      vars:
+        foreman_cloud_connector_url: https://foreman.example.com
+        foreman_cloud_connector_user: admin
+        foreman_cloud_connector_password: changeme
+```
--- a/roles/cloud_connector/defaults/main.yml
+++ b/roles/cloud_connector/defaults/main.yml
@@ -1,0 +1,4 @@
+---
+foreman_cloud_connector_console_redhat_host: cert.cloud.redhat.com
+foreman_cloud_connector_config_file: /etc/rhc/workers/foreman_rh_cloud.toml
+foreman_cloud_connector_validate_certs: true
--- a/roles/cloud_connector/tasks/cleanup_receptor.yml
+++ b/roles/cloud_connector/tasks/cleanup_receptor.yml
@@ -1,0 +1,52 @@
+---
+- name: Identify Satellite ID from Satellite
+  ansible.builtin.uri:
+    url: "{{ foreman_cloud_connector_url }}/api/settings?search=name%20%3D%20instance_id"
+    user: "{{ foreman_cloud_connector_user }}"
+    password: "{{ foreman_cloud_connector_password }}"
+    return_content: true
+    validate_certs: true
+    force_basic_auth: true
+  register: satellite_uuid_json
+
+- name: Find receptor configs
+  find:
+    paths: /etc/receptor
+    file_type: directory
+  register: receptor_configs
+
+- include_tasks: delete_source.yaml
+  loop: "{{ receptor_configs['files'] }}"
+
+- name: Check if foreman-maintain installed
+  ansible.builtin.package_facts:
+    manager: auto
+
+- name: Collect service facts
+  ansible.builtin.service_facts:
+
+- name: Stop any receptor services
+  ansible.builtin.service:
+    state: stopped
+    enabled: false
+    name: "{{ item.name }}"
+  loop: "{{ ansible_facts.services.values() | selectattr('name', 'regex', 'receptor\\@.+.service') | list }}"
+
+- name: Remove receptor packages
+  yum:
+    state: absent
+    name:
+      - receptor
+      - python3-receptor-satellite
+      - ansiblerole-satellite-receptor-installer
+    disable_plugin: foreman-protector
+
+- name: Cleanup receptor config
+  file:
+    state: absent
+    path: /etc/receptor
+
+- name: Remove receptor service file
+  file:
+    state: absent
+    path: /etc/systemd/system/receptor@.service
--- a/roles/cloud_connector/tasks/delete_source.yaml
+++ b/roles/cloud_connector/tasks/delete_source.yaml
@@ -1,0 +1,20 @@
+---
+- name: Get source record
+  ansible.builtin.uri:
+    url: "https://{{ foreman_cloud_connector_console_redhat_host }}/api/sources/v3.1/sources?filter[source_ref]={{ satellite_uuid_json.json.results[0].value }}"
+    client_cert: "{{ item.path }}/cert.pem"
+    client_key: "{{ item.path }}/key.pem"
+    return_content: true
+  register: source_record_response
+
+- name: Delete Source
+  vars:
+    source_id: "{{ source_record_response.json.data[0].id }}"
+  when: "source_record_response.json.data | length == 1"
+  ansible.builtin.uri:
+    url: "https://{{ foreman_cloud_connector_console_redhat_host }}/api/sources/v3.1/sources/{{ source_id }}"
+    client_cert: "{{ item.path }}/cert.pem"
+    client_key: "{{ item.path }}/key.pem"
+    method: "DELETE"
+    status_code:
+      - 204
--- a/roles/cloud_connector/tasks/main.yml
+++ b/roles/cloud_connector/tasks/main.yml
@@ -1,0 +1,62 @@
+---
+- name: Install yggdrasil-worker-forwarder and rhc
+  yum:
+    name:
+      - rhc
+      - yggdrasil-worker-forwarder
+    disable_plugin: foreman-protector
+
+- name: Create workers directory
+  file:
+    state: directory
+    path: "/etc/rhc/workers"
+    owner: root
+    group: root
+    mode: 0755
+
+- name: Configure rhc-cloud-connector-worker
+  ansible.builtin.template:
+    src: "foreman_rh_cloud.toml.j2"
+    dest: "{{ foreman_cloud_connector_config_file }}"
+    owner: root
+    group: root
+    mode: 0640
+
+- name: Create rhcd worker
+  ansible.builtin.copy:
+    dest: /usr/libexec/rhc/foreman-rh-cloud-worker
+    content: |
+      #!/bin/bash
+
+      CONFIG_FILE="{{ foreman_cloud_connector_config_file }}" exec /usr/libexec/yggdrasil-worker-forwarder
+    owner: root
+    group: root
+    mode: 0755
+
+- name: Ensure rhcd started
+  service:
+    name: rhcd
+    state: started
+
+- name: Read client ID from CN of consumer
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
+  register: cert_output
+  when: cert_output is not defined
+
+- name: Set client ID in Satellite
+  ansible.builtin.uri:
+    url: "{{ foreman_cloud_connector_url }}/api/settings/rhc_instance_id"
+    user: "{{ foreman_cloud_connector_user }}"
+    password: "{{ foreman_cloud_connector_password }}"
+    body:
+      setting:
+        value: "{{ client_id }}"
+    method: "PUT"
+    validate_certs: "{{ foreman_cloud_connector_validate_certs }}"
+    force_basic_auth: true
+    body_format: json
+  vars:
+    client_id: "{{ cert_output.stdout.replace('subject= /CN=', '') }}"
+
+- name: Cleanup receptor
+  include: cleanup_receptor.yml
--- a/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
+++ b/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
@@ -1,0 +1,8 @@
+exec = "/usr/libexec/yggdrasil-worker-forwarder"
+protocol = "grpc"
+env = [
+  "FORWARDER_USER={{ foreman_cloud_connector_user }}",
+  "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}",
+  "FORWARDER_HANDLER=foreman_rh_cloud"
+]
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,7 @@ jobs:
           - stable-2.10
           - stable-2.11
           - stable-2.12
+          - stable-2.13
           - devel
         include:
           - python: "2.7"
@@ -35,13 +36,13 @@ jobs:
           - python: "3.10"
             ansible: "devel"
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v2.3.1
+        uses: actions/setup-python@v3
         with:
           python-version: ${{ matrix.python }}
       - name: Restore pip cache
-        uses: actions/cache@v2.1.7
+        uses: actions/cache@v3
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
@@ -50,7 +51,7 @@ jobs:
       - name: Install libyaml-dev for PyYAML
         run: sudo apt-get install -y libyaml-dev
       - name: Install Ansible
-        run: pip install --upgrade git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
+        run: pip install --upgrade https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz
       - name: Install dependencies
         run: make test-setup
       - name: Install required collections for ansible-base (2.10+)
@@ -63,9 +64,9 @@ jobs:
   docs:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v2.3.1
+        uses: actions/setup-python@v3
         with:
           python-version: "3.8"
       - name: Restore pip cache
@@ -88,24 +89,26 @@ jobs:
   lint:
     runs-on: ubuntu-latest
     steps:
-      - uses: actions/checkout@v2
+      - uses: actions/checkout@v3
       - name: Set up Python
-        uses: actions/setup-python@v2.3.1
+        uses: actions/setup-python@v3
         with:
           python-version: "3.8"
       - name: Restore pip cache
-        uses: actions/cache@v2.1.7
+        uses: actions/cache@v3
         with:
           path: ~/.cache/pip
           key: ${{ runner.os }}-pip-${{ hashFiles('**/requirements*.txt') }}
           restore-keys: |
             ${{ runner.os }}-pip-
       - name: Install dependencies
-        run: pip install --upgrade -r requirements-lint.txt ansible
+        run: |
+          pip install --upgrade ansible-core
+          pip install --upgrade -r requirements-lint.txt
       - name: Run lint
         run: make lint
       - name: Upload collection artifact
-        uses: actions/upload-artifact@v2
+        uses: actions/upload-artifact@v3
         with:
           name: collection
           path: theforeman-operations-*.tar.gz
@@ -120,15 +123,16 @@ jobs:
           - stable-2.10
           - stable-2.11
           - stable-2.12
+          - stable-2.13
           - devel
     steps:
       - uses: actions/checkout@v2
       - name: Set up Python
-        uses: actions/setup-python@v2.3.1
+        uses: actions/setup-python@v3
         with:
           python-version: "3.8"
       - name: Install Ansible
-        run: pip install --upgrade git+https://github.com/ansible/ansible.git@${{ matrix.ansible }}
+        run: pip install --upgrade https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz
       - name: Install dependencies
         run: make test-setup
       - name: Run tests
--- a/requirements-dev.txt
+++ b/requirements-dev.txt
@@ -1,5 +1,5 @@
 docker
-rstcheck
+rstcheck==3.3.1 # from https://github.com/ansible/ansible/raw/devel/test/sanity/code-smell/rstcheck.requirements.txt
 cryptography<3.1; python_version < '3.6'
 -r requirements-lint.txt
 -r requirements.txt


roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
```diff
@@ -0,0 +1,5 @@
+---
+foreman: "{{ foreman_url }}"
+token: "{{ foreman_access_token }}"
+username: "{{ foreman_username }}"
+require_ssl: "{{ foreman_require_ssl }}"
```
ShimShtein: According to the PR in the worker, we are not using a separate toml file. This template won't be used

ehelms: Yea this whole thing is old and needs to be dropped from the role.

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+
+- name: Configure rhc-cloud-connector-worker
+  copy:
+    dest: "/etc/rhc/workers/rhc-worker-cloud-connector.toml"
+    content: |
+      directive = 'cloud-connector'
+      env = [
+        "YGG_FORWARDER_USER={{ username }}",
+        "YGG_FORWARDER_PASSWORD={{ password }}",
+        "YGG_FORWARDER_URL={{ url }}"
```
ShimShtein: This will change sooner than later. Yggdrasil does not allow passing ENV variables with prefix `YGG_`

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+  file:
+    state: directory
+    path: "/etc/rhc/workers"
+    owner: root
+    group: root
+    mode: 755
+
+- name: Configure rhc-cloud-connector-worker
+  copy:
+    dest: "/etc/rhc/workers/rhc-worker-cloud-connector.toml"
```
ShimShtein: I think we should use the package config consts here with a default set in case the package config is not set in older version

ShimShtein:
```suggestion
    dest: "/etc/rhc/workers/foreman_rh_cloud.toml"
```
The name of the file will have to be the same as the `directive` used in the code.

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+- name: Create rhcd worker
+  ansible.builtin.copy:
+    dest: /usr/libexec/rhc/rhc-cloud-connector-worker
+    owner: root
+    group: root
+    mode: 0755
+    content: |
+      #!/bin/bash
+
+      CONFIG_FILE=/etc/rhc/workers/rhc-worker-cloud-connector.toml /usr/bin/yggdrasil-worker-forwarder
```
ShimShtein: Can we get the toml file name as a parameter? in case we change it, we can reference it here.

roles/cloud_connector/defaults/main.yml
```diff
@@ -0,0 +1,2 @@
+---
+cloud_connector_config_file: /etc/rhc/workers/rhc-worker-cloud-connector.toml
```
ShimShtein: According to [#6](https://github.com/ShimShtein/yggdrasil-worker-forwarder/pull/6/files#diff-2873f79a86c0d8b3335cd7731b0ecf7dd4301eb19a82ef7a1cba7589b5252261R36), the directive name is derived from the config file name, hence:
```suggestion
cloud_connector_config_file: /etc/rhc/workers/foreman_rh_cloud.toml
```

ShimShtein (changes requested):

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,37 @@
+---
+- name: Check if foreman-maintain installed
+  ansible.builtin.package_facts:
+    manager: auto
+
+- name: Unlock packages
+  command: "foreman-maintain packages unlock"
+  when: "'foreman-maintain' in ansible_facts.packages"
```
evgeni: shouldn't that be `rubygem-foreman_maintain`?

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,37 @@
...
+
+- name: Stop any receptor services
+  command: systemctl stop receptor*
+
+- name: Disable any receptor services
+  command: systemctl disable receptor*
+  ignore_errors: true
```
evgeni: Should we find the services using [service_facts](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/service_facts_module.html) instead and use the `service` module to stop/disable them?

evgeni: Plus, below, you purge `/etc/systemd/system/receptor@.service` but this here would also have stopped/disabled the `receptoral-unified-backup` service (sorry, if your keyboard now contains more coffee than before) ;-)

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+
+    - name: Install yggdrasil-worker-forwarder
+      package:
+        name: yggdrasil-worker-forwarder
+        state: present
+
+    - name: Install rhc
+      package:
+        name: rhc
+        state: present
```
evgeni: is there a reason to do this in two DNF transactions?

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,54 @@
...
+    owner: root
+    group: root
+    mode: 0755
+
+- name: Ensure rhcd started
+  service:
+    name: rhcd
+    state: started
+
+- include: cleanup_receptor.yml
```
evgeni:
```suggestion
- name: Cleanup receptor
  include_tasks: cleanup_receptor.yml
```

ShimShtein: We are missing the RHC id reporting. We can use this API to report it back to Foreman: https://apidocs.theforeman.org/foreman/3.2/apidoc/v2/settings/update.html

roles/cloud_connector/tasks/delete_source.yaml
```diff
@@ -0,0 +1,23 @@
+---
```
ehelms, line 1: This new file handles fetching a source record and issuing a delete to help cleanup the receptor end to end.

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Collect service facts
+  ansible.builtin.service_facts:
+
+- name: Stop any receptor services
+  ansible.builtin.service:
+    state: stopped
+    enabled: false
+    name: "{{ item.name }}"
+  loop: "{{ ansible_facts.services.values() | selectattr('name', 'regex', 'receptor\\@.+.service') | list }}"
```
ehelms, line 33: @evgeni Trying to pull out just receptor services.

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- when: (['receptor', 'python3-receptor-satellite', 'ansiblerole-satellite-receptor-installer'] | difference(ansible_facts.packages.keys()) | list | length) == 0
```
ehelms: Maybe this is unnecessary but keeps from having to do lock and unlock on foreman-maintain.

roles/cloud_connector/templates/rhc-cloud-connector-worker.j2
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+CONFIG_FILE="{{ cloud_connector_config_file }}" /usr/libexec/yggdrasil-worker-forwarder
```
ehelms: This location depends upon -- https://github.com/theforeman/foreman-packaging/pull/7758

roles/cloud_connector/tasks/main.yml
ehelms: I have added this at the bottom of this set of tasks -- take a look at the logic and let me know your thoughts.

roles/cloud_connector/defaults/main.yml
ehelms: Are you wanting this tied to foreman_rh_cloud rather than the cloud connector moniker?

roles/cloud_connector/README.md
```diff
@@ -0,0 +1,30 @@
...
+
+```yaml
+- hosts: target-host
+  roles:
+    - role: theforeman.operations.cloud_connector
+      vars:
+        foreman_cloud_connector_url: https://foreman.example.com
+        foreman_cloud_connector_username: admin
+        foreman_cloud_connector_password: changeme
+        foreman_installer_scenario: katellhangeme'
```
evgeni: these all look wrong ;)

roles/cloud_connector/tasks/cleanup_receptor.yml
evgeni: I  like :)

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
+---
+- name: Find receptor configs
+  find:
+    paths: /etc/receptor
+    file_type: directory
+  register: receptor_configs
+
+- name: Identify Satellite ID from Satellite
```
evgeni, line 2: does this step have to happen before `delete_source.yaml`? It feels wrong to "find receptor configs" first, then get the ID which is totally unrelated, and then clean up stuff for receptor (again).

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+  ansible.builtin.uri:
+    url: "https://{{ foreman_cloud_connector_url }}/api/settings?search=name%20%3D%20instance_id"
```
evgeni: in the example playbook, `foreman_cloud_connector_url` contained the `https://` already (which makes it identical to `foreman_url` in FAM, which I like), so I guess it should be dropped here?

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+    user: "{{ foreman_cloud_connector_username }}"
+    password: "{{ foreman_cloud_connector_password }}"
+    return_content: yes
+    validate_certs: yes
```
evgeni: I'd argue some users will want to override this.

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+
+- name: Cleanup receptor config
+  file:
+    state: absent
+    path: /etc/receptor
+
+- name: Remove receptor service file
+  file:
+    state: absent
+    path: /etc/systemd/system/receptor@.service
```
evgeni, line 52: should this trigger a `systemctl daemon-reload` or similar?

```yaml
- name: systemd reload
  systemd:
    daemon_reload: yes
```

roles/cloud_connector/tasks/cleanup_receptor.yml
evgeni: ah, reading the source of `delete_source.yaml` explains it. then i'd order this: get id, find configs, include delete?

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+    name: rhcd
+    state: started
+
+- name: Read client ID from CN of consumer
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
+  register: cert_output
+
+- name: Extract CN
+  set_fact:
+    client_id: "{{ cert_output.stdout.replace('subject= /CN=', '') }}"
```
evgeni: I recently learned that you can just set vars for a single task, instead of `set_fact` that is global, see below

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,73 @@
...
+    url: "https://{{ foreman_cloud_connector_url }}/api/settings/rhc_instance_id"
+    user: "{{ foreman_cloud_connector_username }}"
+    password: "{{ foreman_cloud_connector_password }}"
+    body:
+      setting:
+        value: "{{ client_id }}"
+    method: "PUT"
+    validate_certs: yes
+    force_basic_auth: yes
+    body_format: json
```
evgeni:
```suggestion
- name: Set client ID in Satellite
  ansible.builtin.uri:
    url: "https://{{ foreman_cloud_connector_url }}/api/settings/rhc_instance_id"
    user: "{{ foreman_cloud_connector_username }}"
    password: "{{ foreman_cloud_connector_password }}"
    body:
      setting:
        value: "{{ client_id }}"
    method: "PUT"
    validate_certs: yes
    force_basic_auth: yes
    body_format: json
  vars:
    client_id: "{{ cert_output.stdout.replace('subject= /CN=', '') }}"
```

roles/cloud_connector/vars/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+foreman_cloud_connector_url: https://foreman.example.com
+foreman_cloud_connector_username: admin
+foreman_cloud_connector_password: changeme
```
evgeni: if they are set in `vars`, they can't be overriden.

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
```diff
@@ -0,0 +1,5 @@
+env = [
+  "FORWARDER_USER={{ satellite_username }}",
+  "FORWARDER_PASSWORD={{ satellite_password }}",
+  "FORWARDER_URL={{ satellite_url }}"
```
evgeni: where do these come from?

evgeni:
I am still very torn on the whole branding/naming situation (which is not specific to this role, but gets exposed here very prominently).

The rhc/rhcd naming will never be true for RHEL-rebuilds and Stream, as it's a RHEL-only thing. On the other hand the yggdrasil naming might be present everywhere.

I wonder if we should put these things into [variable files and load the variables with something like](https://docs.ansible.com/ansible/latest/collections/ansible/builtin/include_vars_module.html):
```yaml
- name: Load a variable file based on the OS type, or a default if not found. Using free-form to specify the file.
  include_vars: "{{ lookup('first_found', params) }}"
  vars:
    params:
      files:
        - 'os-{{ansible_distribution}}.yaml'
        - 'family-{{ansible_os_family}}.yaml'
        - default.yaml
      paths:
        - 'vars'
```

Maybe even include `{{ ansible_distribution_major_version }}` in the names?

roles/cloud_connector/tasks/cleanup_receptor.yml
ehelms: To what end?

roles/cloud_connector/vars/main.yml
ehelms: Oops, I got this mixed up, again.

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
ehelms: :magic_wand:

roles/cloud_connector/defaults/main.yml
```diff
@@ -0,0 +1,7 @@
+---
+foreman_cloud_connector_url: https://foreman.example.com
+foreman_cloud_connector_username: admin
+foreman_cloud_connector_password: changeme
+foreman_cloud_connector_console_redhat_host: cert.cloud.redhat.com
+foreman_cloud_connector_config_file: /etc/rhc/workers/rhc-worker-cloud-connector.toml
```
ShimShtein: Since the directive name is [derived](https://github.com/RedHatInsights/yggdrasil/blob/b35da0f6e66c6fbac58ec9839925bafdddf28b0b/cmd/yggd/worker.go#L42) from configuration file name, we have to call this file `foreman_rh_cloud.toml`.
I know it's inconvenient, and I don't like it either.

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
```diff
@@ -0,0 +1,5 @@
+env = [
+  "FORWARDER_USER={{ foreman_cloud_connector_username }}",
+  "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}"
+]
```
ShimShtein:
```suggestion
exec = “/usr/libexec/yggdrasil-worker-forwarder”
protocol=grpc
]
```
We need to add those according to yggdrasil readme.

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
```diff
@@ -0,0 +1,5 @@
+env = [
+  "FORWARDER_USER={{ foreman_cloud_connector_username }}",
+  "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}"
```
ShimShtein:
```suggestion
  "FORWARDER_URL={{ foreman_cloud_connector_url }}",
  "FORWARDER_HANDLER=foreman_rh_cloud"  
```
We need to specify the handler in case we're not using the batch wrapper (newer yggdrasil)

roles/cloud_connector/tasks/cleanup_receptor.yml
evgeni: so that systemctl doesn't show them anymore?

ShimShtein (changes requested):

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
ehelms: Got a reference? Right now we derive the handler from the file name, you want to just set it it explicitly?

roles/cloud_connector/templates/foreman-rh-cloud-worker.j2
```diff
@@ -0,0 +1,3 @@
+#!/bin/bash
+
+CONFIG_FILE="{{ cloud_connector_config_file }}" /usr/libexec/yggdrasil-worker-forwarder
```
adamruzicka: Since this is just a wrapper to set an env variable, it should probably exec into the the worker to not leave shells running around
```suggestion
CONFIG_FILE="{{ cloud_connector_config_file }}" exec /usr/libexec/yggdrasil-worker-forwarder
```

roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
```diff
@@ -0,0 +1,7 @@
+exec = “/usr/libexec/yggdrasil-worker-forwarder”
```
adamruzicka: Maybe this is just me, but the quotes look like U+201D `“` instead of U+0022 `"`

ehelms: Goood catch.

roles/cloud_connector/templates/rhc-worker-cloud-connector.toml.j2
ShimShtein: We are setting it in the code only if `CONFIG_FILE` is set here: https://github.com/ShimShtein/yggdrasil-worker-forwarder/blob/main/main.go#L36, but if the variable is empty (which would be the case for newer yggd version) we are expecting it to be set explicitly [here](https://github.com/ShimShtein/yggdrasil-worker-forwarder/blob/main/main.go#L47).

Yggdrasil [does something similar](https://github.com/RedHatInsights/yggdrasil/blob/b35da0f6e66c6fbac58ec9839925bafdddf28b0b/cmd/yggd/worker.go#L42), but it uses this for different purposes (like naming the pid file), and does not reuse the information from the worker (the one that we send from the `FORWARDER_HELPER` variable)

roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
ShimShtein: Nice catch! The suggestion was copied from google doc that autocompleted the quotes :facepalm:

roles/cloud_connector/tasks/delete_source.yaml
```diff
@@ -0,0 +1,23 @@
...
+- name: Get source record
+  ansible.builtin.uri:
+    url: "https://{{ foreman_cloud_connector_console_redhat_host }}/api/sources/v3.1/sources?source_ref={{ satellite_uuid_json.json.results[0].value }}"
```
lindgrenj6: The filtering params here just need to be updated,
```suggestion
    url: "https://{{ foreman_cloud_connector_console_redhat_host }}/api/sources/v3.1/sources?filter[source_ref]={{ satellite_uuid_json.json.results[0].value }}"
```

ehelms:
Current state is green. I have not added any new tests dedicated to this role as testing this workflow is complex from a setup perspective. I am open to suggestions.

ehelms:
I did some production testing and pushed some updates I found.

lindgrenj6 (approved):
Looks good from the sources end!

roles/cloud_connector/defaults/main.yml
```diff
@@ -0,0 +1,4 @@
+---
+foreman_cloud_connector_console_redhat_host: cert.cloud.redhat.com
```
evgeni, line 2: in which case would one want to override this?

roles/cloud_connector/README.md
```diff
@@ -0,0 +1,29 @@
...
+Install and configure Red Hat Cloud Connector
+
+Role Variables
+--------------
+
+Required:
+
+- `foreman_url`: The URL of the Foreman server.
+- `foreman_user`: The username cloud connector will use to talk to Foreman API.
+- `foreman_password`: The password cloud connector will use to talk to Foreman API.
```
evgeni: below (and in the playbooks) you use `foreman_cloud_connector_` prefix, I'm fine with either, but it should be consistent ;)

roles/cloud_connector/tasks/delete_source.yaml
```diff
@@ -0,0 +1,23 @@
...
+
+- name: Delete Source
+  ansible.builtin.uri:
+    url: "https://{{ foreman_cloud_connector_console_redhat_host }}/api/sources/v3.1/sources/{{ source_id }}"
+    client_cert: "{{ item.path }}/cert.pem"
+    client_key: "{{ item.path }}/key.pem"
+    method: "DELETE"
+    status_code:
+      - 204
+  when: source_id is defined
```
evgeni:
```suggestion
  vars:
    source_id: "{{ source_record_response.json.data[0].id }}"
  when: "source_record_response.json.data | length == 1"
```

roles/cloud_connector/tasks/delete_source.yaml
```diff
@@ -0,0 +1,23 @@
...
+    client_cert: "{{ item.path }}/cert.pem"
+    client_key: "{{ item.path }}/key.pem"
+    return_content: true
+  register: source_record_response
+
+- name: Set source ID fact
+  set_fact:
+    source_id: "{{ source_record_response.json.data[0].id }}"
+  when: "source_record_response.json.data | length == 1"
+
```
evgeni:
```suggestion
```

roles/cloud_connector/defaults/main.yml
adamruzicka: I think in the past folks were testing sat+receptor against a testing environment in cloud

ehelms: Good question - I would have to ask if there is a staging environment for testing.

roles/cloud_connector/README.md
ehelms: Touche.

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,61 @@
...
+    - name: Unlock packages
+      ansible.builtin.command: "foreman-maintain packages unlock"
+      when: "'rubygem-foreman_maintain' in ansible_facts.packages"
+
+    - name: Remove receptor packages
+      package:
+        name:
+          - receptor
+          - python3-receptor-satellite
+          - ansiblerole-satellite-receptor-installer
```
evgeni:
```suggestion
      yum:
        name:
          - receptor
          - python3-receptor-satellite
          - ansiblerole-satellite-receptor-installer
        disable_plugin: foreman-protector
```

this should make the whole lock/unlock dance go away ;)

roles/cloud_connector/defaults/main.yml
evgeni: fair :)

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
+---
+- name: Install yggdrasil-worker-forwarder and rhc
+  yum:
+    state: absent
```
evgeni: *absent*?

roles/cloud_connector/tasks/main.yml
```diff
@@ -0,0 +1,63 @@
...
+    name: rhcd
+    state: started
+
+- name: Read client ID from CN of consumer
+  ansible.builtin.command: openssl x509 -in /etc/pki/consumer/cert.pem -subject -noout
+  register: cert_output
+  when: cert_output is not defined
+
+- name: Set client ID in Satellite
+  ansible.builtin.uri:
```
evgeni, line 47: if only we had a setting module in some collection… ;)

roles/cloud_connector/tasks/cleanup_receptor.yml
```diff
@@ -0,0 +1,52 @@
+---
+- name: Identify Satellite ID from Satellite
+  ansible.builtin.uri:
```
evgeni, line 3: if only we had an setting_info module in some collection… ;)

evgeni (commented):
ansible-wise this reads good to me (besides the absent/present issue for the rhc package).

we could think of making this depend on the other collection, but I am not forcing you to ;)

roles/cloud_connector/tasks/main.yml
ehelms: Ooops... I guess I was on a role... :fireworks: :laughing:

roles/cloud_connector/tasks/main.yml
ehelms: Somebody should write that

ehelms:
> we could think of making this depend on the other collection, but I am not forcing you to ;)

If I had more than this one use case for this collection I'd be happy to, otherwise I hesitate. Granted, Ansible should just handle dependency resolution and everything nicely for us right? :smile:

roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
```diff
@@ -0,0 +1,7 @@
+exec = "/usr/libexec/yggdrasil-worker-forwarder"
+protocol = "grpc"
+env = [
+  "FORWARDER_USER={{ foreman_cloud_connector_user }}",
+  "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}"
```
ShimShtein: We still need to add 
```suggestion
  "FORWARDER_URL={{ foreman_cloud_connector_url }}",
  "FORWARDER_HANDLER=foreman_rh_cloud"
```
for new versions of yggdrasil

ShimShtein (changes requested):

roles/cloud_connector/tasks/main.yml
mdellweg: _Holding up the irony sign._

https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/modules/setting.py

roles/cloud_connector/tasks/cleanup_receptor.yml
mdellweg: https://github.com/theforeman/foreman-ansible-modules/blob/develop/plugins/modules/setting_info.py

ehelms: The ole double dose of sarcasm.

evgeni (approved):
ack from an ansible PoV, I have no idea if the thing does what it says, but that's for someone else to worry ;)

PR #60, opened 2022-02-17
enable foreman and katello modules on el8

evgeni:


roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -18,13 +18,40 @@
...
+    warn: False
+    creates: /etc/dnf/modules.d/katello.module
+  when:
+    - foreman_repositories_katello_version is defined
+    - foreman_repositories_katello_version == "nightly" or foreman_repositories_katello_version is version('4.4', '>=')
+    - ansible_distribution_major_version == '8'
+
 - when:
     - foreman_repositories_katello_version is defined
     - ansible_distribution_major_version == '8'
```
ekohl: Shouldn't this be scoped to `< 4.4`? At least, `pki-core` (and possibly powertools`) shouldn't be needed anymore, right?

evgeni: pki-core true, powertools are still needed (thats a repo, not a module)

but I wanted to keep the diff small.

evgeni:
```
Error: Problems in request:", "missing groups or modules: katello:el8
```

:thinking:

evgeni:
fascinating, the 3.2 repos really don't have module metadata on the CDN, but whyyyy
(they do on koji)

evgeni:
https://github.com/theforeman/foreman-infra/blob/f847559002fc18bffc759b190a712aa934a60af0/puppet/modules/web/files/deploy-yumrepo.sh#L50

:crying_cat_face:

roles/foreman_repositories/tasks/redhat.yml
evgeni: implemented now

ehelms (approved):

PR #61, opened 2022-02-17
switch to using stream8 containers, now that 8 is EOL

evgeni:


evgeni:
This fixes some things, but not all, and nightly installation of Katello is broken right now anyways, so the test results are not exactly meaningful.

Merging as is, yell at me another day.

PR #62, opened 2022-02-23
run tests with a script and report *which* roles failed

evgeni:


tests/molecule.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
+
+failed_roles=()
+
+for roledir in roles/*/molecule; do
+  role_path=$(dirname $roledir)
+  role_name=$(basename $role_path)
+  pushd ${role_path}
+  molecule test --all || failed_roles+=($role_name)
+  popd
```
ekohl, line 10: You can try to create groups for easier debugging: https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#grouping-log-lines
```suggestion
  echo "::group::Ansible role $role_name"
  pushd ${role_path}
  molecule test --all || failed_roles+=($role_name)
  popd
  echo "::endgroup::"
```

ekohl (approved):
I manually tested the bash logic and that appears to be correct.

I also thought about using [errors](https://docs.github.com/en/actions/using-workflows/workflow-commands-for-github-actions#setting-an-error-message) but I'm not sure how to relate it to a file.

tests/molecule.sh
evgeni: Oh, that's nifty!

evgeni: Molecule already does grouping, if it detects GHA, and nested groups seem to be broken: https://github.com/actions/runner/issues/802

I'll play with that idea in a follow-up.

PR #63, opened 2022-02-23
add GHA group markers

evgeni:
as suggested by Ewoud


evgeni:
https://github.com/theforeman/foreman-operations-collection/runs/5307871885?check_suite_focus=true

Screenshot from 2022-02-23 19-13-09

The folding is off, as nested groups aren't supported (yet).

@ekohl do you think this adds to the readability of the output, or not?

ekohl (approved):
I had the exact question myself. It can be useful to have the marker for the role so I'm leaning to merging this.

PR #65, opened 2022-02-23
don't complain about no changed when for upgrade command

evgeni:
it's in a sensible block

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -113,7 +113,6 @@ jobs:
   molecule:
     runs-on: ubuntu-latest
     strategy:
-      max-parallel: 4
       matrix:
         ansible:
           - stable-2.9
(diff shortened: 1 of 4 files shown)


evgeni:
well, at least that role fails later now :woman_shrugging:

evgeni:
with all those changes, the remaining failing role is… postgresql_upgrade

it fails because it needs to install foreman 2.3 but 2.3 can't be (by default) deployed on CentOS8 stream as the used PostgreSQL puppet module doesn't recognize it :(

ehelms:
Originally we wanted to use an installer run to feel more confident about an upgrade with "real" data. Older Foreman versions will start to wane as this indicates so I'd be in favor of dropping the installer part of the postgresql upgrade role and simplifying the testing.

evgeni:
> Originally we wanted to use an installer run to feel more confident about an upgrade with "real" data. Older Foreman versions will start to wane as this indicates so I'd be in favor of dropping the installer part of the postgresql upgrade role and simplifying the testing.

So, like this? Setup a 10, upgrade to 12?

evgeni:
:green_apple:

ehelms (approved):

PR #66, opened 2022-02-24
switch to Foreman 3.1 by default

evgeni:


ekohl (approved):
Looks like some failures could be solved by picking https://github.com/theforeman/foreman-packaging/commit/fea6988c9b6dfb20f0d2423509e385f733cfa086 to 3.1.

evgeni:
They could indeed.

PR #67, opened 2022-02-25
Foreman-installer task made to be async and apt cache fixes

kyle-williams-1:
Foreman-installer is a long running task and is prone to connection timeouts. It's recommended to make these async (ref. https://docs.ansible.com/ansible/latest/user_guide/playbooks_async.html) in order to avoid those connection timeouts. I've tested this and it works well. `changed_when` and `failed_when` continue to work as expected.

I've also included some fixes related to hanging of apt tasks on Ubuntu.

### DIFF
--- a/roles/installer/defaults/main.yml
+++ b/roles/installer/defaults/main.yml
@@ -5,3 +5,4 @@ foreman_installer_verbose: true
 foreman_installer_no_colors: false
 foreman_installer_options: []
 foreman_installer_locale: 'en_US.UTF-8'
+foreman_installer_timeout: 1800
--- a/roles/puppet_repositories/tasks/debian.yml
+++ b/roles/puppet_repositories/tasks/debian.yml
@@ -1,10 +1,20 @@
 ---
-- name: 'Install GPG package'
+- name: Install GPG package
   package:
-    name: 'gpg'
+    name: gpg
     state: present
     update_cache: true
 
-- name: 'Install Puppet repository'
+- name: Install Puppet repository
   apt:
-    deb: https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb
+    deb: "https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb"
+    state: present
+  register: puppet_apt_deb_install
+
+- name: Update apt cache if necessary
+  apt:
+    update_cache: true
+    force_apt_get: true
+  when: puppet_apt_deb_install.changed
+  tags:
+    - skip_ansible_lint  # skip lint since we cannot use handler, and cache must be updated immediately
--- a/roles/installer/README.md
+++ b/roles/installer/README.md
@@ -18,6 +18,7 @@ Optional:
 - `foreman_installer_no_colors`: Disables color output from the installer
 - `foreman_installer_command`: Installer command to run, can be used by derivative projects to specify a branded command
 - `foreman_installer_locale`: Locale to run the installer with, this must not be ```C```, defaults to ```en_US.UTF-8```
+- `foreman_installer_timeout`: Time allowed for installer to run before it is assumed failed, defaults to 30m
 
 Example Playbooks
 -----------------
--- a/roles/installer/tasks/main.yml
+++ b/roles/installer/tasks/main.yml
@@ -31,7 +31,7 @@
     name: "{{ foreman_installer_package }}"
     state: present
 
-- name: 'Run installer'
+- name: Run installer
   command: >
     {{ foreman_installer_command }}
     --scenario {{ foreman_installer_scenario }}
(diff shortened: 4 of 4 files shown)


kyle-williams-1:
Hello. Just wondering if I might be able to get this reviewed. Included are must have fixes in order to install Foreman.

evgeni:
To understand this correctly, you're running the role via a remote connection, and waiting for the installer to finish sometimes produces timeouts, whereas using async does not, as that triggers re-checks every now and then?

The timeout of 300 (5 minutes) seems rather low, especially when running katello scenario, but otherwise I can follow the logic, I think ;)

roles/puppet_repositories/tasks/debian.yml
```diff
@@ -1,10 +1,16 @@
...
+- name: Install Puppet repository
   apt:
-    deb: https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb
+    deb: "https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb"
+    state: present
+  register: puppet_apt_deb_install
+
+- name: Update apt cache if necessary
+  apt: update_cache=yes force_apt_get=yes
+  when: puppet_apt_deb_install.changed
```
evgeni: this should be probably a handler, notified by the previous task?

roles/puppet_repositories/tasks/debian.yml
```diff
@@ -1,10 +1,16 @@
...
-- name: 'Install Puppet repository'
+- name: Install Puppet repository
   apt:
-    deb: https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb
+    deb: "https://apt.puppet.com/puppet{{ foreman_puppet_repositories_version }}-release-{{ ansible_distribution_release }}.deb"
+    state: present
+  register: puppet_apt_deb_install
+
+- name: Update apt cache if necessary
+  apt: update_cache=yes force_apt_get=yes
```
ehelms: Please prefer the nested yaml style than inline approach for declaring parameters:

```suggestion
  apt:
    update_cache: yes
    force_apt_get: yes
```

ehelms:
@kyle-williams-1 We merged some test fixes that will force you to rebase this PR.

roles/puppet_repositories/tasks/debian.yml
kyle-williams-1: Totally agree; however, there is an issue with using a handler here. Handlers run at the end of a play unless we use the `meta: flush_handlers` sometime after the repo is added and before the installer is called. Generally speaking, I think most people using this repo are adding the puppet repos and then calling the installer. If we were to flush_handlers between the two, we might force a bunch of other handlers to run that the developer did not intend to be run before the play completed.

The foreman-installer must not update the repo cache before installing the package. Nonetheless, this change will ensure we update the repo prior to using the installer.

roles/puppet_repositories/tasks/debian.yml
kyle-williams-1: Agreed! Updated.

kyle-williams-1:
> To understand this correctly, you're running the role via a remote connection, and waiting for the installer to finish sometimes produces timeouts, whereas using async does not, as that triggers re-checks every now and then?
>
> The timeout of 300 (5 minutes) seems rather low, especially when running katello scenario, but otherwise I can follow the logic, I think ;)

That's exactly right! The async will keep the session alive. I've updated the timeout to 10m. Let me know if that sounds reasonable.

kyle-williams-1:
I've done a rebase and addressed the comments. Let me know if  I can do anything else. Thanks for the review!

roles/puppet_repositories/tasks/debian.yml
evgeni: Ah, yes of course!

kyle-williams-1: I just saw the pipeline failed due to the lint error. I've added `skip_ansible_lint` tag. Let me know if you think this is overkill or there is a better way to handle this. :)

kyle-williams-1:
I see katello install failing in the ci tests, and I wonder if this is due to the default timeout I've added of 10m. I'll attempt and increase to 30m. Do we have a more reasonable timeout?

evgeni:
Yeah, it is/was failing due to the timeout.

I'm afraid there is no good baseline here. 30m is certainly in the ballpark of "that's how long a installer runs for katello" :/

@ehelms ideas/suggestions how to aproach this?

ehelms:
I can live with 30 minutes, that ought to be a good value without being too extreme or too low. We may find through use in CI and other environments we need to up this but the parameter gives us that control.

roles/installer/defaults/main.yml
```diff
@@ -5,3 +5,4 @@ foreman_installer_verbose: true
 foreman_installer_no_colors: false
 foreman_installer_options: []
 foreman_installer_locale: 'en_US.UTF-8'
+foreman_installer_timeout: 1800
```
ehelms, line 8: Can you add this to the documentation here -- https://github.com/theforeman/foreman-operations-collection/blob/develop/roles/installer/README.md?plain=1#L13

kyle-williams-1: Updated 👍

ehelms:
@evgeni fixed some CI issues in his PR that I just merged, so I think if you rebase and push the tests should get past the pip part

ehelms (approved):
@evgeni take a look too please

evgeni (approved):

evgeni:
Hmm. One thing that this did break is checkmode:

```
TASK [Run installer] ******************************************************************
fatal: [foreman]: FAILED! => {"changed": false, "msg": "check mode and async cannot be used on same task."}
```

Is there a good way around this?

evgeni:
let's try https://github.com/theforeman/foreman-operations-collection/pull/77

PR #70, opened 2022-03-16
pin ansible-lint to pre-6

evgeni:
ansible-lint 6 wants ansible-core >= 2.12, which breaks testing with
older ansibles


evgeni:
(split out of #60, as the changes are technically unrelated)

mdellweg (approved):

PR #77, opened 2022-03-25
don't fail installer role in check mode

evgeni:


mdellweg (approved):
I remember using these workarounds for test playbooks. But seems like we need this here now.

evgeni:
If only CI would now also run…

PR #84, opened 2022-04-27
don't collect package information facts, we don't need them

evgeni:


ehelms:
Oh, right! You found a smarter way to do this.

ehelms (approved):

PR #85, opened 2022-05-05
Fix FORWARDER_URL in foreman_rh_cloud.toml.j2

jeremy-audet:
See: https://issues.redhat.com/browse/SAT-10187

### DIFF
--- a/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
+++ b/roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
@@ -3,6 +3,6 @@ protocol = "grpc"
 env = [
   "FORWARDER_USER={{ foreman_cloud_connector_user }}",
   "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
-  "FORWARDER_URL={{ foreman_cloud_connector_url }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}/api/v2/rh_cloud/cloud_request",
   "FORWARDER_HANDLER=foreman_rh_cloud"
 ]


roles/cloud_connector/templates/foreman_rh_cloud.toml.j2
```diff
@@ -3,6 +3,6 @@ protocol = "grpc"
 env = [
   "FORWARDER_USER={{ foreman_cloud_connector_user }}",
   "FORWARDER_PASSWORD={{ foreman_cloud_connector_password }}",
-  "FORWARDER_URL={{ foreman_cloud_connector_url }}",
+  "FORWARDER_URL={{ foreman_cloud_connector_url }}/api/v2/rh_cloud/cloud_request",
```
ehelms, line 6: @ShimShtein can you verify?

evgeni: Sounds like something the worker should figure out on its own to me?

jeremy-audet: Maybe this fix is wrong, and `foreman_cloud_connector_url` should be set to `https://$(hostnamectl --static)/api/v2/rh_cloud/cloud_request` by whatever/whoever is calling the playbook?

evgeni: No, the `foreman_cloud_connector_url` is the URL to the "root" of the Foreman Installation, as it's also used in other places in the playbook to call the Foreman API.

ehelms: > Sounds like something the worker should figure out on its own to me?

What do you mean? The worker is a generic worker that knows how to forward to a URL.

evgeni: > > Sounds like something the worker should figure out on its own to me?
> 
> What do you mean? The worker is a generic worker that knows how to forward to a URL.

Oh yeah, right, it's not foreman specific. Yeah, then here is the right place

ShimShtein:
ACK from my side. I have missed this from the original round.

PR #87, opened 2022-05-11
Add HTTP proxy support to Cloud Connector role

ehelms:


roles/cloud_connector/tasks/main.yml
```diff
@@ -58,5 +58,9 @@
   vars:
     client_id: "{{ cert_output.stdout.replace('subject= /CN=', '') }}"
 
+- name: Configure HTTP proxy
```
evgeni, line 61: At this step, rhcd is already running and nothing restarts it.

Also, nothing triggers a systemd daemon reload, so the dropin is not loaded either.

evgeni (changes requested):
Is this a proxy for talking to cloud or to Foreman?

ehelms:
> Is this a proxy for talking to cloud or to Foreman?

Good question - RHC talking to cloud. However, given this a generic solution that affects `net`, I can see how this might cause talking to Foreman to route through the proxy too which is not what we want (https://github.com/ShimShtein/yggdrasil-worker-forwarder/blob/main/server.go#L60). @ShimShtein thoughts on if this would be a problem and how we could prevent it?


This way of configuring comes from -- https://github.com/RedHatInsights/yggdrasil/wiki/FAQ#how-do-i-connect-to-a-broker-using-an-httphttps-proxy

ShimShtein:
@ehelms according to the [go docs](https://pkg.go.dev/net/http#Client) the `net` library supports also the `no_proxy`/`NO_PROXY` env variable. We can add an exception for the foreman server there.
On a deeper thought, it seems that the request would always go to `localhost` anyway, since the RHCd would be installed on the same machine as Satellite. This makes the `no_proxy` rule much more stable. I don't suppose there are many cases where you would want to add a proxy to access local machine.

evgeni:
`localhost` is not part of the cert and not what we setup as the foreman url
(Also, that would break Ansible)

ShimShtein:
So no localhost shortcut :) We still can add the foreman url to the no_proxy list.

evgeni:
Yepp, that part sounds reasonable. Just wanted to point out that `localhost` is a bad idea :)

roles/cloud_connector/templates/proxy.conf.j2
```diff
@@ -0,0 +1,3 @@
+[Service]
+Environment=HTTPS_PROXY={{ foreman_cloud_connector_http_proxy }}
+Environment=NO_PROXY={{ foreman_cloud_connector_url }}
```
ehelms, line 3: @evgeni @ShimShtein added NO_PROXY here, this is what I interpreted by the meaning of how we want to use it here to prevent proxying between calls to Foreman

evgeni: looks correct to me

roles/cloud_connector/tasks/http_proxy.yml
```diff
@@ -0,0 +1,22 @@
...
+    src: "proxy.conf.j2"
+    dest: /etc/systemd/system/rhcd.service.d/proxy.conf
+    owner: root
+    group: root
+    mode: 0640
+
+- name: Restart and reload RHC service
+  ansible.builtin.systemd:
+    name: rhcd
+    state: restarted
```
evgeni: technically, it only needs restarting if the dropin changed, but oh well…

ehelms: Lets see if I handled handlers correctly

evgeni: It does read correct to me

evgeni (approved):

PR #92, opened 2022-05-19
Fix client_id parsing for openssl on EL8

ehelms:


evgeni:
Can't we just use  `openssl x509 -in cert.pem -noout -subject -nameopt nofname`?

No idea if that works on EL7.

Edit: man page of openssl 1.0.2 suggested it does, no box to test with at hand.

With that, we'd just need to strip the `subject=` prefix. No CN at all.

ehelms:
That's a **little** bit cleaner, but still not congruent... because of course.

EL8
```
openssl x509 -in /etc/pki/tls/cert.pem -noout -subject -nameopt nofname
subject=ACCVRAIZ1, PKIACCV, ACCV, ES
```

EL7:
```
openssl x509 -in /etc/pki/tls/cert.pem -noout -subject -nameopt nofname
subject= ACCVRAIZ1, PKIACCV, ACCV, E
```

Notice the tiny space after the `=`

evgeni:
`-nameopt nofname,sep_comma_plus` maybe?

Edit: nope

ehelms:
Yea, I cannot find any option that normalizes across the versions of openssl.

evgeni (approved):
whatever, there is a nicer way, for sure, somewhere, but I am not going to search for it

PR #95, opened 2022-06-08
release 1.2.2

evgeni:


changelogs/changelog.yaml
```diff
@@ -45,3 +45,10 @@ releases:
       bugfixes:
       - fix parsing of openssl certificate subject for EL7 and EL8
     release_date: '2022-05-23'
+  1.2.2:
+    changes:
+      bugfixes:
+      - cloud_connector role - use a regular expression to parse the CN from the cert
+    fragments:
+    - 94-subject-regex.yml
```
ehelms, line 53: Something appears to have actually deleted the fragment in this PR.

evgeni: Yes, which is fine. The fragment is only needed for the generation of the changelog, not afterwards.

evgeni: This line says: we used that fragment.

As you can configure antsibull not to delete, but then you need to track which are already used.

ehelms: Ohh... that's weird.

ehelms (approved):

PR #100, opened 2022-08-24
Enable rhcd.service

jeremy-audet:
Ensure `rhcd.service` starts after system reboots.

Fix: https://issues.redhat.com/browse/SAT-12447


evgeni:
LGTM.

@ehelms what do you think?

ehelms (approved):

PR #101, opened 2022-08-29
add backup role

evgeni:


roles/backup/templates/foreman-backup.sh.j2
```diff
@@ -0,0 +1,22 @@
+#!/bin/bash -eu
```
evgeni, line 1: I do wonder if this should be a Python/Ruby script instead.

The "find old full backups" logic is rather hard to follow imho.

ehelms: Python would be easier to follow. I wondered if this should have two modes of operation? setting up a cronjob and doing it in a more Ansible native way? aka the Ansible role itself doing the backup allowing users to run it from Tower or Jenkins or some other system?

evgeni: I think before going that route, I'd write a "Backup configuration file" support for f-maintain.

ehelms: > I think before going that route, I'd write a "Backup configuration file" support for f-maintain.

Would you expand a little on what that would look like? And how it then helps?

evgeni: So today, we have `foreman-maintain` that on the one hand knows the application and thus "what to backup" (filesystem paths, db dumps, etc), but also contains a slim layer on top of `tar` to make storage of backups "better" (by allowing incrementals etc).

I am not a big fan of the later, as it makes FM a backup tool (see requests for LVM snapshot functionality etc), which I think it should not be. This is partially also the fault of the app (you need db and disk storage to "match" to have proper backup).

Now, said that, we *are* a backup tool today. As we can create those backups, but we make a poor job at managing them.

Like, when I want to create an incremental backup, I first need to manually find its ancestor.
And there is no way to prune old backups.
Yet, all this is functionality of a real backup tool.

So for example, instead of having this code to do an incremental backup:
```bash
  LAST=$(ls -td -- $DESTINATION/*/ | head -n 1)
  foreman-maintain backup offline --assumeyes --incremental "$LAST" $DESTINATION
```
I would expect people to just be able to say 
```bash
foreman-maintain backup offline --assumeyes --incremental $DESTINATION
```
and FM finds the correct ancestor itself.

Same for pruning, instead of the weird `grep` and `find` gore I've done now, I would expect to be able to say:
```bash
foreman-maintain backup prune --assumeyes --before <date> $DESTINATION
```
or
```bash
foreman-maintain backup prune --assumeyes --keep-full 2 $DESTINATION
```

Again, I don't think that we should have a such sophisticated backup command, but if we do, it should be complete.

ehelms (approved):

PR #110, opened 2022-11-24
Update ansible-lint to 6.1+ and enable only-builtins rule

evgeni:


evgeni:
The Python fixes are in https://github.com/theforeman/foreman-operations-collection/pull/111, but irrelevant for the change here.

ehelms (approved):

PR #120, opened 2023-03-23
make CI happy

evgeni:


evgeni:
I have no idea what's up with molecule, but the issues are totally unrelated to the changes in this (or the other) PRs :sob:

evgeni:
cool cool

I can't reproduce this with `quay.io/centos/centos:stream8@660b13416363` base container, but I can with `quay.io/centos/centos:stream8@f0422dc107eb`

today is CI day, huh?

evgeni:
okay, so there seem to be some issue with systemd and I don't get it as it does work just fine locally here, (but I am using Podman, not Docker)

evgeni:
okay, I *think* I know what happens. systemd takes a moment to start after the container is up and running, and that takes longer on docker and thus we end up in a state where things don't work :exploding_head:

ekohl (approved):

PR #123, opened 2023-04-04
don't mark proxy.conf systemd drop-in word-inaccessible

evgeni:
there is no benefit and systemd warns about this:

    systemd[1]: Configuration file /etc/systemd/system/rhcd.service.d/proxy.conf is marked world-inaccessible.
    This has no effect as configuration data is accessible via APIs without restrictions. Proceeding anyway.

### DIFF
--- a/changelogs/fragments/bz2169682-world-inaccessible.yml
+++ b/changelogs/fragments/bz2169682-world-inaccessible.yml
@@ -1,0 +1,2 @@
+minor_changes:
+  - cloud_connector role - don't mark proxy.conf systemd drop-in word-inaccessible, there is no benefit and systemd warns about this (https://bugzilla.redhat.com/show_bug.cgi?id=2169682)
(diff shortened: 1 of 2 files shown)


evgeni:
```
FileNotFoundError: [Errno 2] No such file or directory: '/var/cache/dnf/download_lock.pid'
```

what fresh hell is this now?!

ehelms:
huh, why does systemd care my drop-in file permissions? What if I need to put information in that file that I don't want everyone on the system to be able to read?

evgeni:
That's exactly the point. Everyone *can* read it, regardless of permissions, by using `systemctl cat rhcd.service`. So there is no protection whatsoever by the file permissions and thus one should not use any that make people think non root can't read it.

ehelms:
> That's exactly the point. Everyone _can_ read it, regardless of permissions, by using `systemctl cat rhcd.service`. So there is no protection whatsoever by the file permissions and thus one should not use any that make people think non root can't read it.

:facepalm:

ehelms (approved):

PR #131, opened 2023-09-01
disable var-naming[no-role-prefix] and force a var pattern instead

evgeni:


evgeni:
the failing tests are unrelated and the fix for those can be found in https://github.com/theforeman/foreman-operations-collection/pull/132

ehelms (approved):

PR #137, opened 2023-11-17
Drop support for Foreman pre 3.2 / Katello pre 4.4 repositories

evgeni:


evgeni:
The devel failure is unrelated and being fixed in #138

ehelms (approved):

PR #139, opened 2023-11-20
Drop PostgreSQL upgrade role

evgeni:


ehelms (approved):
Here's to hoping we don't need this again.

evgeni:
> Here's to hoping we don't need this again.

Even if we do, we'll have to change things, so :shrug:

PR #140, opened 2023-11-20
add metrics role

evgeni:


roles/metrics/README.md
```diff
@@ -0,0 +1,25 @@
+theforeman.operations.metrics
```
evgeni, line 1: should this be called `metrics` or `monitoring`? the guide is called monitoring, but it's really more like "metrics"

roles/metrics/defaults/main.yml
```diff
@@ -0,0 +1,11 @@
+---
+foreman_metrics_url: "https://{{ ansible_fqdn | default('localhost') }}/metrics"
+foreman_pcp_additional_packages: "{{ ['pcp-system-tools'] if ansible_os_family == 'RedHat' else [] }}"
+foreman_pcp_optional_agents_redhat:
+  - apache
+  - openmetrics
+  - postgresql
+  - redis
+foreman_pcp_optional_agents: "{{ foreman_pcp_optional_agents_redhat if ansible_os_family == 'RedHat' else [] }}"
```
evgeni: this currently means we do not enable those PMDAs on Debian at all, but in reality we only want not to install their packages… it's complicated.

roles/metrics/README.md
```diff
@@ -0,0 +1,25 @@
...
+=============================
+
+Setup Performace Co-Pilot and Grafana to grather metrics of a Foreman installation.
```
ehelms:
```suggestion
Setup Performace Co-Pilot and Grafana to gather metrics of a Foreman installation.
```

roles/metrics/README.md
```diff
@@ -0,0 +1,25 @@
...
+
+This role is heavily based on the [performancecopilot.metrics](https://github.com/performancecopilot/ansible-pcp) collection.
+
+Role Variables
+--------------
+
+* `foreman_metrics_url`: The URL of the OpenMetrics endpoint of Foreman, defaults to `https://{{ ansible_fqdn | default('localhost') }}/metrics`.
```
ehelms, line 11: is the endpoint `/metrics` actually changeable?

roles/metrics/README.md
```diff
@@ -0,0 +1,25 @@
...
+* `foreman_pcp_optional_agents`: The optional PCP agents to enable, defaults to `[apache, openmetrics, postgresql, redis]`.
+* `foreman_metrics_grafana_enabled`: Whether or not Grafana should be installed and configured on the system, defaults to `true` on Red Hat family systems, and to `false` otherwise.
+* `foreman_metrics_grafana_pmproxy_url`: The URL of the `pmproxy` service to be used as the Grafana datasource, defaults to `http://{{ ansible_fqdn | default('localhost') }}:44322`.
```
ehelms, line 14: Does this port need to be mentioned in our ports guide with the addition of our monitoring guide?

roles/metrics/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+foreman_metrics_url: "https://{{ ansible_fqdn | default('localhost') }}/metrics"
+foreman_pcp_additional_packages: "{{ ['pcp-system-tools'] if ansible_os_family == 'RedHat' else [] }}"
```
ehelms: Do these `foreman_pcp` violate the naming convention rules? Should they be `foreman_metrics_pcp`?

roles/metrics/README.md
ehelms: The role itself feels like it's about metrics. And metrics can be combined into monitoring. So yes, I think you keep this name.

ehelms:
How generic are the Grafana steps? Do users with an external Grafana need to do the same steps to get our setup to work?

roles/metrics/README.md
evgeni: Do we mention the Candlepin 23443 port? It's a local connection (unless you do remote-grafana, but then you'd usually also have the pmproxy remotely), so not really needed.

roles/metrics/defaults/main.yml
evgeni: Technically yeah.

We do configure `ansible-lint` to accept everything that is `foreman_` prefixed tho, but would be good style to include the role name.
https://github.com/theforeman/foreman-operations-collection/blob/9c78cf04186e24e00b424d5ac4eba03b99f454bb/.ansible-lint#L6

roles/metrics/README.md
evgeni: In Foreman? no.

But you could use this role to monitor a remote Foreman.

evgeni:
> How generic are the Grafana steps? Do users with an external Grafana need to do the same steps to get our setup to work?

There are various definitions of "remote grafana" (only grafana remote, or pmlogger/pmproxy/grafana remote, the later is more common in PCP world), but I do think the steps should be generic enough (didn't try out tho).

roles/metrics/README.md
ehelms: Where I was going with it was, should we only allow users to configure the hostname rather than the full URL if the `/metrics` part is not configurable?

roles/metrics/README.md
ehelms: We don't, but we specifically set Candlepin to localhost. This appears that it would set it to the hostname (so I assume the firewall would not incorporate this as a in and out connection). Will pmproxy always be remote if Grafana is remote?

roles/metrics/defaults/main.yml
```diff
@@ -0,0 +1,12 @@
+---
+foreman_metrics_url: "https://{{ ansible_fqdn | default('localhost') }}/metrics"
```
ehelms, line 2: In what cases is `ansible_fqdn` not set?

roles/metrics/README.md
evgeni: Technically they also have the choice of http vs https (if they disable https enforcement)

roles/metrics/defaults/main.yml
evgeni: gather_facts: false

roles/metrics/README.md
evgeni: For the firewall/kernel it would be always considered a local connection, regardless of the IP/hostname (if it's on the same host, obviously).

I think in most deployments, pmlogger/pmproxy/grafana would be on one box, which might or might not be the same box as Foreman. There is however no technical reason not to host them each on different boxen.

https://access.redhat.com/documentation/en-us/red_hat_enterprise_linux/8/html/monitoring_and_managing_system_status_and_performance/setting-up-pcp_monitoring-and-managing-system-status-and-performance#pcp-deployment-architectures_setting-up-pcp has a few examples

ehelms (approved):

PR #141, opened 2023-11-21
cleanup molecule prepare playbooks

evgeni:
the mixture of `roles`, `pre_tasks`, `tasks` and `post_tasks` is confusing. just use `tasks` and when a role is needed `include_role`.

### DIFF
--- a/roles/backup/molecule/redhat/prepare.yml
+++ b/roles/backup/molecule/redhat/prepare.yml
@@ -2,9 +2,7 @@
 - name: Prepare
   hosts: all
   gather_facts: true
-  pre_tasks:
+  tasks:
     - name: Install crontabs
       ansible.builtin.package:
         name: crontabs
-  roles:
-    - backup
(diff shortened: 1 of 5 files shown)


roles/backup/molecule/redhat/prepare.yml
```diff
@@ -2,9 +2,7 @@
 - name: Prepare
   hosts: all
   gather_facts: true
-  pre_tasks:
+  tasks:
     - name: Install crontabs
       ansible.builtin.package:
         name: crontabs
-  roles:
-    - backup
```
Griffin-Sullivan, line 10: Do you not still need this backup role to run?

evgeni: not in "prepare", it's still there in "converge":
https://github.com/theforeman/foreman-operations-collection/blob/466fc476b0baa9b45a3efaaa484750215d7482a7/roles/backup/molecule/redhat/converge.yml#L6

it was actually a mistake that it was mentioned here. "prepare" should be just the setup steps, not the actual role execution

Griffin-Sullivan: makes sense

Griffin-Sullivan (approved):

PR #149, opened 2024-03-05
only enable powertools for Katello < 4.10

evgeni:
4.10+ has no katello-agent support and no need for powertools

### DIFF
--- a/roles/foreman_repositories/tasks/redhat.yml
+++ b/roles/foreman_repositories/tasks/redhat.yml
@@ -36,6 +36,8 @@
 - name: Enable powertools for Katello
   when:
     - foreman_repositories_katello_version is defined
+    - foreman_repositories_katello_version != 'nightly'
+    - foreman_repositories_katello_version is version('4.10', '<')
     - ansible_distribution_major_version == '8'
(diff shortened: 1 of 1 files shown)


roles/foreman_repositories/tasks/redhat.yml
```diff
@@ -36,6 +36,8 @@
 - name: Enable powertools for Katello
   when:
     - foreman_repositories_katello_version is defined
+    - foreman_repositories_katello_version != 'nightly'
+    - foreman_repositories_katello_version is version('4.10', '<')
```
Griffin-Sullivan, line 40: Do you have to specify not nightly here?

Griffin-Sullivan (approved):

roles/foreman_repositories/tasks/redhat.yml
evgeni: `version()` uses Python's version comparison code, and that does not cope well with versions that are not "numeric", so we filter out `nightly` earlier, so `version()` is never called with it.

ekohl (approved):

PR #151, opened 2024-03-13
Add Puppet 7 and 8 to the list of supported releases

evgeni:


roles/puppet_repositories/vars/main.yml
```diff
@@ -1,3 +1,5 @@
 foreman_puppet_repositories_supported_versions:
   - "5"
```
ekohl, line 2: This was dropped from yum.puppet.com so should this be dropped? Perhaps in a separate PR.

ekohl (approved):

roles/puppet_repositories/vars/main.yml
evgeni: Yeah.

evgeni: https://github.com/theforeman/foreman-operations-collection/pull/153

PR #155, opened 2024-03-26
always require foreman_repositories_version to be set

evgeni:


Griffin-Sullivan (approved):
LGTM. How do these get tested? I can't decipher the `/tests` folder

evgeni:
The 5000 ft view: [Molecule](https://ansible.readthedocs.io/projects/molecule/en/latest/) is a tool that (using Ansible playbooks) can setup a test node and then execute tests against this node. The tests are usually a `converge` playbook that is run twice and a/ is not allowed to error out and b/ in the second run not allowed to produce changes, and a `verify` playbook that then has some asserts based on the work `converge` was supposed to be doing.
In this PR I only touched converge for Debian, as for RH we were already setting the version (no idea why) and the asserts remain the same.

ehelms (approved):

PR #159, opened 2024-04-10
add description metadata to all roles

evgeni:


evgeni:
old molecule will be happy once https://github.com/theforeman/foreman-operations-collection/pull/160 is in

ehelms (approved):

PR #161, opened 2024-04-27
run molecule in parallel

evgeni:


evgeni:
if you wanna know the gory details, read https://www.die-welt.net/2024/04/running-ansible-molecule-tests-in-parallel/
otherwise: this speeds up the test from 26+ minutes to 13 minutes. enjoy!

.github/workflows/main.yml
```diff
@@ -146,10 +146,15 @@ jobs:
       - name: Install Ansible 2.9 compatible molecule and Jinja
         run: pip install --upgrade 'molecule[docker]<4.0' 'Jinja2<3.1'
         if: matrix.ansible == 'v2.9.17'
+      - name: Install collection dependencies
+        run: cd roles/backup && molecule dependency -s debian
```
ekohl: Looking at https://ansible.readthedocs.io/projects/molecule/usage/ I wonder if `-s` should be written as `--scenario-name`.

ekohl (approved):

.github/workflows/main.yml
evgeni: How dare you taking my "I DEMAND LONGOPTS EVERYWHERE" and point it at me?! :)

ekohl: It's Monday, it shouldn't be easy today.

PR #168, opened 2024-05-31
Switch testing to CentOS Stream 9

### LINKED ISSUE #167: Switch foreman-operations-collection to testing with Alma

evgeni:
Fixes: #167


roles/foreman_repositories/molecule/redhat/molecule.yml
```diff
@@ -4,14 +4,10 @@ dependency:
...
 platforms:
-  - name: centos8
-    image: quay.io/centos/centos:stream8
-    command: /sbin/init
-    tmpfs:
-      - /run
-      - /tmp:exec,mode=777
-    volumes:
-      - /sys/fs/cgroup:/sys/fs/cgroup:ro
+  - name: almalinux8
```
evgeni, line 7: this is the only role that differentiates between el8 and el9, so I opted for testing on both

roles/foreman_proxy_certs_generate/molecule/default/Dockerfile.j2
```diff
@@ -0,0 +1,22 @@
...
+{% if item.env is defined %}
+{% for var, value in item.env.items() %}
+{% if value %}
+ENV {{ var }} {{ value }}
+{% endif %}
+{% endfor %}
+{% endif %}
+
+RUN if [ $(command -v apt-get) ]; then export DEBIAN_FRONTEND=noninteractive && apt-get update && apt-get install -y python3 sudo bash ca-certificates iproute2 python3-apt aptitude systemd systemd-sysv && apt-get clean && rm -rf /var/lib/apt/lists/*; \
+    elif [ $(command -v dnf) ]; then dnf makecache && dnf --assumeyes install /usr/bin/python3 /usr/bin/python3-config /usr/bin/dnf-3 sudo bash iproute systemd && dnf clean all; \
```
evgeni: the default cs9 container doesn't have systemd installed, so I took the Dockerfile.j2 from molecule and added it.

roles/backup/molecule/redhat/molecule.yml
```diff
@@ -4,14 +4,8 @@ dependency:
 driver:
   name: ${DRIVER_NAME:-podman}
 platforms:
-  - name: centos8
-    image: quay.io/centos/centos:stream8
-    command: /sbin/init
```
evgeni, line 9: most roles do not need working init, so I dropped it where possible

ekohl (commented):
Would it make sense to store the Dockerfile.j2 in some central place and use symlinks to avoid duplication?

evgeni:
I don't understand why Ansible 2.10 fails

evgeni:
> I don't understand why Ansible 2.10 fails

I still do not, but using an older Molecule fixed it, so I am not going to complain.

ekohl (approved):

PR #169, opened 2024-06-07
allow disabling pulling /metrics endpoint

evgeni:
useful when you want to deploy metrics for a proxy only setup


pablomh:
Makes sense.

ehelms (approved):

PR #175, opened 2025-02-12
Run CI workflows on Ubuntu 24.04 & drop Ansible 2.9-2.11 support

evgeni:


mdellweg (approved):

.github/workflows/main.yml
```diff
@@ -139,10 +139,8 @@ jobs:
...
         with:
           python-version: ${{ matrix.python }}
-      - name: Install Ansible
-        run: pip install --upgrade https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz
-      - name: Install dependencies
-        run: make test-setup
+      - name: Install Ansible and dependencies
+        run: pip install --upgrade https://github.com/ansible/ansible/archive/${{ matrix.ansible }}.tar.gz -r requirements-dev.txt
       - name: Install Ansible 2.9/2.10 compatible molecule and Jinja
         run: pip install --upgrade 'molecule[podman]<4.0' 'Jinja2<3.1'
```
mdellweg: If at all possible, I think you should merge this pip install call too. My observation is that any subsequent pip call will break the previously installed packages without an opportunity to opt out.

I'd suggest to create a constraints files (instead of requirements files).

Something like:
```yaml
- name: "Configure constraints for Ansible <2.11"
  run: |
    echo "molecule<4.0" >> constraints-dev.txt
    echo "Jinja2<3.1" >> constraints-dev.txt
  if: matrix.ansible == 'v2.9.17' || matrix.ansible == 'stable-2.10'
- name: "Install Ansible and dependencies"
  run: |
    touch constraints-dev.txt
    pip install [...] -c constraints-dev.txt
```

evgeni: Sure, or I just drop all that "old ansible" cruft and let it rot in hell :)

PR #196, opened 2026-03-18
Release 4.0.0

evgeni:


evgeni:
Ansible devel is failing because of https://github.com/ansible/ansible/pull/86090, but I'd prefer not to fix it right now

shubhamsg199 (approved):

PR #198, opened 2026-06-16
Fix openvox_repositories on Ubuntu

evgeni:


evgeni:
`devel` fixes in https://github.com/theforeman/foreman-operations-collection/pull/199

